PR #13, opened 2021-06-15
Responsive Sidebar > Responsive Header Nav in Narrow Viewports

andlum:
## What?
Fixed a few usability issues with the narrow view of the Docs:
- Added a `Back to Radar.io` button to the sidebar - this is always visible
- Added a Dashboard link to the sidebar - this is visible only in narrow views
- Moved responsive sidebar button to top-left (in a hacky way... more in the `How` section)

## Why?
There are some usability issues with the current Navbar + Sidebar in narrower viewports:
- Page has 2 hamburger menus (but no clear indication which menu opens what)
- Default responsive sidebar button obfuscates scroll to top button

This change makes the sidebar menu the main navigation for the page (since its more contextually relevant to the content anyway).

## How?
This is definitely a hackier implementation since the responsive sidebar menu just obfuscates the original menu with CSS. I went with this route mostly for the speed of implementation, but this is definitely a code smell.

A better implementation would be:
1.) Adding a context for Sidebar, e.g. `SidebarContext`
2.) Updating `SidebarContext` via the `DocSidebar` swizzled component
3.) Updating the Navbar component to ingest data from the `SidebarContext` and replacing the button entirely

But this'll take considerable effort and may not play well with future updates since the `Navbar` is not a [swizzle-friendly component yet](https://docusaurus.io/docs/lifecycle-apis#getswizzlecomponentlist) and this requires rewriting rather than just adding elements

## Screenshots (optional)
Wide view
image

Narrow view
image

Narrow view w/ sidebar modal
image

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -234,6 +234,21 @@ strong {
   height: auto;
 }
 
+/* Sidebar overrides */
+
+.menu__list-item {
+  margin: 0;
+}
+
+.menu__link {
+  border-radius: 0;
+}
+
+.menu__link--active,
+.menu__link--active:hover {
+  font-weight: var(--ifm-font-weight-semibold);
+  background: var(--ifm-menu-color-background);
+}
 
 /* Table of Contents overrides */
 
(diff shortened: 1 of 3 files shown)


andlum:
https://app.clubhouse.io/radarlabs/story/6481/convert-docs-mobile-sidebar-into-a-modal

PR #18, opened 2021-06-17
Add Segment to Docs

andlum:

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -9,6 +9,15 @@ module.exports = {
   favicon: 'img/favicon.png',
   organizationName: 'radarlabs', // Usually your GitHub org/user name.
   projectName: 'docs', // Usually your repo name.
+  plugins: [
+    [
+      'docusaurus-plugin-segment',
+      {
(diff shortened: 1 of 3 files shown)


andlum:
https://app.clubhouse.io/radarlabs/story/6526/add-segment-to-docs

docusaurus.config.js
```diff
@@ -9,6 +9,15 @@ module.exports = {
   favicon: 'img/favicon.png',
   organizationName: 'radarlabs', // Usually your GitHub org/user name.
   projectName: 'docs', // Usually your repo name.
+  plugins: [
+    [
+      'docusaurus-plugin-segment',
+      {
+        apiKey: 'DotTOuAVALzcuOiCqq1kPZCRAfGdSa3K'
```
jsani-radar, line 16: are we fine putting our `apiKey` in plaintext here?

jsani-radar (commented):
wondering if this is the right approach given anyone can view this repo and create PRs

docusaurus.config.js
andlum: Yep! This is a client key so theoretically anyone that goes to `radar.io` would be able to read this key from inspecting our DOM

PR #20, opened 2021-06-18
adjust font sizes on mobile, and some other small adjustments

kochis:
## What?
Mobile adjustments:
* Reduce font sizes on mobile viewports
* Set close-icon weight to `normal`
* Set desktop navbar to `hidden` when mobile button is overlaid to prevent it from leaking through

### Font Fixes
image

### Navbar Fixes
image

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -104,8 +104,8 @@ h6 {
 
 /* Responsive font adjustments */
 @media (max-width: 996px) {
-  h1, h1.docTitle_src-theme-DocItem- {
-    font-size: 2.5rem;
+  h1 {
+    font-size: 2.25rem;
   }
   h2 {
     font-size: 2rem;
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -247,12 +247,18 @@ strong {
 }
 
 @media (max-width: 996px) {
+  /* Offset header height*/
+  body {
+    padding-top: 60px;
+  }
+  /* Use a fixed (instead of sticky) header */
+  .navbar--fixed-top {
+    position: fixed;
+  }
+
+  /* Hide the drawer icon thats being overlaid */
   .navbar__toggle {
     visibility: hidden;
-  }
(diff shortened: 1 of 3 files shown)


src/css/custom.css
```diff
@@ -230,6 +246,16 @@ strong {
...
 }
 
+@media (max-width: 996px) {
+  .navbar__toggle {
+    visibility: hidden;
+  }
+  .sidebarMenuIcon_src-theme-DocSidebar-.sidebarMenuCloseIcon_src-theme-DocSidebar- {
+    font-weight: normal;
+    margin-top: -2px;
+  }
```
andlum, line 256: This might be better to make this change in the swizzled component https://github.com/radarlabs/documentation/blob/main/src/theme/DocSidebar/styles.module.css#L109-L118

src/css/custom.css
```diff
@@ -102,6 +102,22 @@ h6 {
   letter-spacing: 1px;
 }
 
+/* Responsive font adjustments */
+@media (max-width: 996px) {
+  h1, h1.docTitle_src-theme-DocItem- {
```
andlum, line 107: {nit} 

The `.docTitle` is auto-generated so it can conceivably change in the future without something explicit on our end (e.g. updating Docusaurus). Adding it [here](https://github.com/radarlabs/documentation/blob/main/src/theme/DocItem/styles.module.css#L8-L11) should be more resilient.

Something like:
```css
.docTitle {
  font-size: 3rem;
}

@media (max-width: 996px) {
  .docTitle {
    font-size: 2.5rem;
  }
}
```

andlum (approved):
Mostly looks good! Let me know if you want to fix the nit or not

src/css/custom.css
andlum: (I kinda wish Docusaurus used the infima variables more strictly so we wouldn't have to do this selective styling but oh well...)

kochis:
I bumped the `h1` down to `2.25rem` to account for even smaller devices (iPhone 5/SE)

image

andlum:
Nice! :shipit:

PR #24, opened 2021-06-28
setting up CODEOWNERS for documentation

tjulien:
## What?
Sets up a `CODEOWNERS` file so that PMs own approving documentation PRs

## Why?
So we can start to spread around the company who can approve documentation


andlum (approved):
Looks good to me!

PR #26, opened 2021-07-15
remove-ad-id-info

brettguenther:
## What?

Remove ad id info from branch documentation.

## Why?

Ad Id is not required for the branch integration (link on device ids) so omit from integration documentation


nickpatrick (approved):
Thanks, @brettguenther. Change looks good.

@andlum Does it matter that `package-lock.json` is changed in this PR?

andlum:
@nickpatrick: Ideally, we shouldn't be including those changes because it changes package versions, but I think these changes should be okay.

PR #33, opened 2021-08-02
Place settings api

brettguenther:
## What?

Document our API to manage place settings.

## Why?

This API is not currently documented.

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1712,3 +1712,120 @@ An event represents a change in user state. Events can be uniquely referenced by
 To create an event, call the [track](#track) endpoint to trigger an event. For example, specify a `latitude` and `longitude` to enter a geofence that you have created.
 
 On iOS and Android, use the [SDK](/sdk) to track users.
+
+
+### Place Settings
+
+Place settings represents your current project configuration for [Places](/places) monitoring. 
+
+###### Object fields
+
+- **`chainMetadata`** (dictionary): Current metadata associated with each Radar chain.
+- **`chainMappings`** (dictionary): Current mapping associated with each Radar chain.
+- **`placeFilters`** (dictionary): Current chains and category filtering for place entry and exit event generation.
+
+#### Get Place Settings
+
+Get current place settings.
+
+###### Definition
+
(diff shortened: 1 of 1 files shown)


docs/api.mdx
```diff
@@ -1712,3 +1712,120 @@ An event represents a change in user state. Events can be uniquely referenced by
...
+###### Sample response
+
+```json
+{
+  "meta": {
+    "code": 200
+  },
+  "chainMetadata": {
+    "mcdonalds": {
+      "availableDeals": false
```
jsani-radar, line 1819: should this be `true` based on the PATCH body? or I guess given the GET above maybe the body should have `availableDeals: false`?

jsani-radar (approved):
small comment, LGTM otherwise!

docs/api.mdx
brettguenther: good catch

docs/api.mdx
```diff
@@ -1712,3 +1712,120 @@ An event represents a change in user state. Events can be uniquely referenced by
...
+
+```json
+{
+    "meta": {
+        "code": 200
+    },
+    "chainMetadata": {
+        "mcdonalds": {
+            "availableDeals": true,
+            "category": "qsr"
```
jkao, line 1761: this `category` attribute might be a bit misleading to the reader. it's just a metadata key

jkao: we can just leave it out

docs/api.mdx
```diff
@@ -1712,3 +1712,120 @@ An event represents a change in user state. Events can be uniquely referenced by
...
+        },
+        "dunkin-donuts": {
+            "availableDeals": true,
+            "category": "breakfast"
```
jkao, line 1765: same as above

docs/api.mdx
jsani-radar: we could also rename it, something like `type`?

jkao: that still might be misleading - if you take a look at the PATCH request below, it's only sending up availableDeals so that might be the most consistent

jsani-radar:
Closed in favor of https://github.com/radarlabs/documentation/pull/39

PR #34, opened 2021-08-03
Update Android continuous description to match SDK docs

nickpatrick:
To match https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/RadarTrackingOptions.kt#L329

We can set `trackingOptions.foregroundService = RadarTrackingOptionsForegroundService()` by default for `CONTINUOUS` in `3.2.x` and add detailed docs for `RadarTrackingOptionsForegroundService` in the future. In the meantime, `RadarTrackingOptionsForegroundService` is documented here https://radarlabs.github.io/radar-sdk-android/sdk/io.radar.sdk/-radar-tracking-options/-radar-tracking-options-foreground-service/index.html

### DIFF
--- a/docs/sdk/tracking.mdx
+++ b/docs/sdk/tracking.mdx
@@ -88,7 +88,7 @@ Option | Efficient | Responsive | Continuous
 
 ### Android presets
 
-- **`RadarTrackingOptions.CONTINUOUS`**: Updates about every 30 seconds while moving or stopped. Moderate battery usage. Starts a foreground service and shows a notification during tracking.
+- **`RadarTrackingOptions.CONTINUOUS`**: Updates about every 30 seconds while moving or stopped. Should be used with a foreground service. Moderate battery usage.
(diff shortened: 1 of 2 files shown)


docs/sdk/tracking.mdx
```diff
@@ -88,7 +88,7 @@ Option | Efficient | Responsive | Continuous
 
 ### Android presets
 
-- **`RadarTrackingOptions.CONTINUOUS`**: Updates about every 30 seconds while moving or stopped. Moderate battery usage. Starts a foreground service and shows a notification during tracking.
+- **`RadarTrackingOptions.CONTINUOUS`**: Updates about every 30 seconds while moving or stopped. Should be used with a foreground service. Moderate battery usage.
```
jsani-radar, line 91: do we want to include language here that we're efficient about battery usage with this preset too?

jsani-radar (approved):
lgtm! thanks for making this small but important change

docs/sdk/tracking.mdx
nickpatrick: We could. This used to say `High battery usage`, I think `Moderate` is more accurate. We can probably go into more detail somewhere else in the docs later on if needed.

PR #38, opened 2021-08-19
Update urban-airship.mdx

ahwreck:
Airship event property is incorrect. See event screenshot.

Screen Shot 2021-08-19 at 11 03 07 AM
Screen Shot 2021-08-19 at 11 02 00 AM


jsani-radar (approved):
great catch, thanks so much!

PR #40, opened 2021-08-20
Auto-generate Radar chains and categories

jkao:
This provides the mechanism to auto-generate some of our docs, such as chains.

We'll also want to do this for groups and categories, but I'd like to see what you guys think about the approach first.

We make use of https://ejs.co/ as the templating mechanism, but I'm open to other approaches, such as Mustache https://mustache.github.io/#demo.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "schemas"]
```
jkao, line 1: it's worth noting that this is not available publicly

jkao:
Added placeholders for you to try out the experience, @kochis 

image

Basically, just follow the README and run `npm run regenerate-templated-docs`

kochis (approved):
LGTM 👍

PR #43, opened 2021-08-23
tracking options foreground service

brettguenther:
## What?

Outline foreground service tracking option use.

## Why?

Currently not documented at this level of detail.

### DIFF
--- a/docs/sdk/tracking.mdx
+++ b/docs/sdk/tracking.mdx
@@ -114,3 +114,13 @@ Option | Efficient | Responsive | Continuous
 **`syncGeofencesLimit`** | `10` | `10` | `0`
 **`foregroundService`** | `null` | `null` | `null`
 **`beacons`** | `false` | `false` | `false`
+
+### Android foreground service tracking options
+
+- **`text`** (string): Sets the text of the notification. Defaults to "Location tracking started".
+- **`title`** (string): Sets the title of the notification. 
+- **`icon`** (integer): Sets the notification icon using a drawable resource id. Defaults to the application icon.
+- **`updatesOnly`** (boolean): Sets whether the foreground should be only displayed after while processing a location update or be persistent.
+- **`activity`** (string): Sets the activity to be launched on interaction with the notification.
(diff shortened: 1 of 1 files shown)


brettguenther:
@jsani-radar - for friday reviews :) - debated on the data types inclusion. If valuable probably makes sense to mimic it for other tracking options + make sure we have all defaults documented.

nickpatrick:
Thanks @brettguenther. I was actually planning to get to this as part of SDK 3.2 docs. This looks solid, just a couple bits of feedback on first glance:
- I might change the headline to `Android foreground service options`. Alternatively or in addition, we might want to mention the class name `RadarTrackingOptionsForegroundService` somewhere.
- The default for notification importance is currently `android.app.NotificationManager.IMPORTANCE_DEFAULT` instead of high: https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/RadarTrackingOptions.kt#L274 https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/RadarForegroundService.kt#L51
- Looks like we're deviating from the SDK docs a bit: https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/RadarTrackingOptions.kt#L247 We should prob try to keep things consistent. If we think some of these edits make the options more understandable, I can incorporate them into the SDK 3.2 PR.
- I like the idea of including data types like `string`. Do we also want to include optional/required? General point that we should probably to keep type names and formatting consistent across the docs. I also wonder if moving all of the tracking options into a table (instead of bullet points) is more readable. Same applies to API params.

jsani-radar:
how to populate `foreground` has been a consistent question from customers using React Native - think we should put links to this section in the Android / RN SDK docs. We might even want a new section like "Background tracking for workforce / always on"

jsani-radar:
where did we end up on this PR @brettguenther? do we want to merge the foreground service options as is or think about larger structural changes?

jsani-radar:
Seems to be already present on the page - https://radar.io/documentation/sdk/tracking#android-foreground-service-options

PR #44, opened 2021-08-23
Update api.mdx

ahwreck:
Update PATCH Trips to Secret authentication

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1401,7 +1401,7 @@ curl "https://api.radar.io/v1/trips/299" \
 
 ###### Authentication level
 
-Publishable
+Secret
 
 ###### Sample response
 


ahwreck:
actually looks like its publishable in the API code. shouldn't this be secret?

https://github.com/radarlabs/server/blob/4107ce191fa731f90fc88f9927e5595bc110aca0/server/routes/apiV1Router.ts#L182

kochis:
I don't think so, since this is intended to be called from the client

ahwreck:
ok - in Hathway's case w/ SnS they are calling it from middleware (NomNom) to set the scheduledArrivalAt timestamp as well. guess to allow it to be called from the client keeping it publishable makes sense.

ahwreck:
do we want to include all the other object fields you can potentially PATCH on a trip as well in the docs?

jsani-radar:
what's the status of this PR @ahwreck? did you want to add the PATCH params or should it be closed?

ahwreck:
No changes needed

PR #51, opened 2021-09-23
Missing and new Amplitude Integration attributes

brettguenther:
## What?

Added a new foreground attribute to Amplitude events and other existing attributes that were missing

## Why?

Because there were gaps in documentation


nickpatrick:
Thanks, @brettguenther. Looks great. Only nit would be to put backticks around the boolean values `true`

brettguenther:
> Thanks, @brettguenther. Looks great. Only nit would be to put backticks around the boolean values `true`

good catch. just changed.

jsani-radar (approved):

PR #64, opened 2021-11-01
[WIP] API reference breakdown

jsani-radar:
TODO:
- [ ] Clean up API reference overview page
- [ ] Fix refs to API links in the rest of the docs

## What?

Before:
Screen Shot 2021-10-31 at 11 43 40 PM

After:
Screen Shot 2021-10-31 at 11 43 54 PM

## Why?

Our API reference section is quite long and can be hard to grok. This has been on my backlog for a very long time, so I took a crack at tackling it as part of fixit. @samfader also relayed similar feedback of making the specific API endpoints more prominent in the documentation structure.
image


nickpatrick:
Nice, I'm a fan of this change

samfader (approved):
@jsani-radar I'm too new to comment on the content itself (and also not sure if any of that changed), but the links all seem to work properly and I really love this new layout.

My only feedback (and I don't think this would require immediate change, but maybe down the line) is that some of the API pages, like geofences and trips, have top-level summaries/descriptions while others, like geocoding and search, do not. It might be nice to have some more context for those APIs (what's the difference between geocoding and search? when might I use one over the other) down the road.

PR #69, opened 2021-11-03
Olo integration docs refresh

brettguenther:
## What?

Update our Olo integration docs

## Why?

Instructions are dated


jsani-radar:
do we need the package updates? seems like they're tangential to this PR's purpose

brettguenther:
> do we need the package updates? seems like they're tangential to this PR's purpose

we do not. reverted package.json

jsani-radar:
content looks solid, but still seeing changes to `package-lock.json`, probably propagated from the earlier `package.json` changes - seems like we'll need a `git rm --cached` or `git reset` on that file https://stackoverflow.com/a/38001958

brettguenther:
> content looks solid, but still seeing changes to `package-lock.json`, probably propagated from the earlier `package.json` changes - seems like we'll need a `git rm --cached` or `git reset` on that file https://stackoverflow.com/a/38001958

thanks @jsani-radar - think i have it reverted now so should hopefully be good for friday docs review.

jsani-radar (approved):

PR #89, opened 2022-01-20
Update android tracking for trips

brettguenther:
## What?

- Remove pseudocode for creating your own foreground service.
- Include reference on customization of the SDK's built in foreground service.
- change .io to .com references on android page

## Why?

Docs are stale

PR #91, opened 2022-01-21
remove explicit stopForegroundService() call

jsani-radar:
To be consistent with removing the `startForegroundService()` call from https://github.com/radarlabs/documentation/pull/89

The Radar SDK now manages the starting / stopping of a foreground service in continuous mode


brettguenther (approved):
Looks good!

PR #92, opened 2022-01-26
Adding configs for Vercel cutover

bschoeneweis:
## What?
- Cleaning up hardcoded urls to use the `baseUrl`
- Adding `vercel.json`

## Why?
Enabling deployment on Vercel.


bschoeneweis:
https://app.shortcut.com/radarlabs/story/8512/spin-up-the-documentation-site-on-vercel

docs/sdk/android-3.1.mdx
```diff
@@ -591,8 +592,8 @@ Your receiver should implement the following:
 
 <div className="alert alert-info">
   To listen for events server-side instead, add a{" "}
-  <a href="../integrations/webhooks">webhook</a> or enable an{" "}
-  <a href="../integrations">integration</a>.
```
kochis, line 595: If we change the `../` to `/` do we still need `useBaseUrl`?

bschoeneweis: Following [this guide from Docusaurus on static assets](https://docusaurus.io/docs/static-assets#referencing-your-static-asset), there is a quote:

> In Markdown, things are different: you can stick to use absolute paths because Docusaurus actually handles them as `require` calls instead of URLs when parsing the Markdown.

In docusaurus, the `require` calls prepend the `baseUrl`.  So I think not having the `useBaseUrl` will have it resolve from root instead of from the `build` directory.

I also didn't love the relative URL incase the structure were to ever change, this seemed to be a more precise definition for the links that use actual html `<a>` elements.

docs/integrations.mdx
```diff
@@ -92,7 +92,7 @@ To filter event types sent to an integration, select or unselect event types und
 
 To filter keys from the request body sent to an integration, add keys, comma-separated, to *Request Key Blacklist*. For example, to skip sending `radar_geofence_ids` to the Segment `identify` call and to skip sending `geofence_id` to the Segment `track` call, add `radar_geofence_ids,geofence_id` to *Request Key Blacklist*.
 
-To filter events sent to an integration based upon a [Geofence](/documentation/geofences) tag, add tags, comma-separated to *Geofence Tag Blacklist*. For example, to skip sending all events to the Amplitude integration for geofences tagged `gas-station` and `cafe`, add `gas-station,cafe` to *Geofence Tag Blacklist*.
+To filter events sent to an integration based upon a [Geofence](/geofences) tag, add tags, comma-separated to *Geofence Tag Blacklist*. For example, to skip sending all events to the Amplitude integration for geofences tagged `gas-station` and `cafe`, add `gas-station,cafe` to *Geofence Tag Blacklist*.
```
jsani-radar, line 95: just want to confirm that this will still navigate to `/documentation/geofences` rather than the old `/geofences` route that goes to the dashboard - I guess that'll mean that this redirect still respects the `baseUrl`

jsani-radar (commented):
lgtm overall - just one small q around link / redirect behavior

docs/integrations.mdx
bschoeneweis: Correct.  So the rule with docusaurus is that for any markdown link `[text](link)`, the `baseUrl` will be prepended automatically.  Just ran a sanity check to confirm as well.

The redirect that's being added is basically to account for the `baseUrl` on Vercel (a bit of a workaround is needed) and shouldn't have any effect on the current production docs since they'll still be on GitHub pages for the time being.

jsani-radar: awesome, context switching to dashboard login is a bit jarring unless it's really obvious, so just want to make sure these links stay internal - thanks

docs/sdk/android-3.1.mdx
kochis: From what I can tell, using `href=/integrations` and `href={useBaseUrl('/integrations')}`, will result in the same thing (docusaurus will prepend the baseUrl to all absolute URLs).

I'm wondering if it's easier to just use the leading `/` everywhere instead of `useBaseUrl`, so that way it's more consistent with how it's done in markdown links. cc @jsani-radar for thoughts

bschoeneweis: Quote here from the docs (https://docusaurus.io/docs/static-assets#in-markdown):

> Docusaurus will only parse links that are in Markdown syntax. If your asset references are using the JSX tag `<a>` / ``, nothing will be done.

kochis: huh, I just tried that locally and is prepended the `/documentation` correctly

bschoeneweis: Hmmm quick way to find out.  I can make the change here and push up to the branch and see if it prepends on the vercel build since that's what important.  I definitely like the consistency of just the absolute paths, but my assumption was that it would break without the `baseUrl`, but only one way to find out

bschoeneweis: Yeah it seemed to work fine without the `useBaseUrl`.  Guess that's why docusaurus is still in beta technically? 🤷‍♂️

jsani-radar: yeah, I slightly prefer sticking to the usual markdown link structure as well, that way it's easier for others to onboard to contributing and one fewer thing to think about

andlum (approved):
This looks good; thanks for doing this!

PR #95, opened 2022-02-03
update background modes

brettguenther:
## What?

Make background modes iOS documentation more prominent and accurate

## Why?

The current documentation is not clear on which tracking modes require this change. Additionally, it is nested under permissions and could easily be missed by developers.

### DIFF
--- a/docs/sdk/ios.mdx
+++ b/docs/sdk/ios.mdx
@@ -317,7 +317,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CLLocationManagerDelegate
 
 ### Background modes
 
-If you are planning to leverage background tracking, such as responsive or continuous modes, in your project settings, go to _Signing & Capabilities_, add _Background Modes_, and turn on _Background fetch_ and _Location updates_. Note that this requires additional justification during App Store review. Learn more [below](#submit-to-app-store).
+If you are planning to leverage background tracking, such as responsive or continuous mode, in your project settings, go to _Signing & Capabilities_, add _Background Modes_, and turn on _Background fetch_ and _Location updates_. Note that this requires additional justification during App Store review. Learn more [below](#submit-to-app-store).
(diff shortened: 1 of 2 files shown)


docs/sdk/ios.mdx
```diff
@@ -317,8 +317,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CLLocationManagerDelegate
   usage strings to your <code>Info.plist</code> file. Finally, check your device
   settings to make sure you haven't already granted location permissions.
 </div>
+    
+### background modes
 
-If you are requesting background permissions, in your project settings, go to _Signing & Capabilities_, add _Background Modes_, and turn on _Background fetch_ and _Location updates_. Note that this requires additional justification during App Store review. Learn more [below](#submit-to-app-store).
+If planning to leverage responsive or continuous background tracking go to your project settings, _Signing & Capabilities_, add _Background Modes_, and turn on _Background fetch_ and _Location updates_. Note that this requires additional justification during App Store review. Learn more [below](#submit-to-app-store).
```
jrtibbetts, line 323: Needs a comma here:

> background tracking, go to

jsani-radar:
@brettguenther do we need to add a corresponding section to Android? right now the headers / sections are the same for both, so this would result in an inconsistency

brettguenther:
closing in favor of: https://github.com/radarlabs/documentation/pull/99

brettguenther:
> @brettguenther do we need to add a corresponding section to Android? right now the headers / sections are the same for both, so this would result in an inconsistency

@jsani-radar - there isn't a corresponding native configuration needed on android specifically for background tracking. options would be:

1. we push this down to background tracking sections rather than in initial config
2. remove the header if we want consistency (there are other differences in sections between android and ios docs)

jsani-radar (approved):

PR #107, opened 2022-03-01
Migrate capacitor to registerPlugin

jsani-radar:
Changes the initialization code for Capacitor to have `registerPlugin` initialize the SDK using the publishable key from `strings.xml` rather than passing the key in manually to `initialize`

### DIFF
--- a/docs/sdk/capacitor.mdx
+++ b/docs/sdk/capacitor.mdx
@@ -59,21 +59,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


phil-brown (changes requested):
Can you also document how to use the old mechanism? I think Nick had mentioned a need for some apps to potentially want to still use `Radar.initialize` after, for example, a login. I added another key in `strings.xml`, such that you can specify to ignore the warning of not initializing this way. Also, if apps are using the old mechanism for initialization, we should note somewhere that the old mechanism of calling initialize will not work when upgrading, since the RadarReceiver stuff is not registered - this is a breaking change.

jsani-radar:
closing as we punted on this feature /cc @nickpatrick

PR #109, opened 2022-03-04
Move to new alert component

matthewisabel:
## What?
I want to move plan support to an alert at the top of feature pages so I went down a rabbit hole first of componentizing all alerts before finding the current examples. This is potentially of dubious value if we want to abandon.

## Why?
Got lost in the refactor

## How?
Swapped some components around

## Screenshots (optional)
Should look the same!

## Anything Else? (optional)
Part 1 of part 2 but I need to cleanup a few things here first


matthewisabel:
@andlum wdyt

andlum (approved):
Oooo! Thank you for taking care of this 🙏 

My only concern is that the props differ from the `Alert` component in [frontend](https://github.com/radarlabs/frontend/blob/master/components/Alert.jsx) but... I'm not sure the `Alert` in Dashboard has a very settled props "API" either so I think this is good for now

src/components/Alert.jsx
```diff
@@ -0,0 +1,33 @@
...
+  
+      if (alertType === "success") {
+        return (
+          <div className={`alert ${styles.alert} ${styles.alertSuccess}`} role="alert">
+            {children}
+          </div>
+        );
+      }
+  
+    return '';
```
andlum, line 30: This is a false-y value so it'll prevent rendering but it's not quite as idiomatic as `return null`

see: https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

src/components/Alert.module.css
```diff
@@ -0,0 +1,61 @@
+/* TODO: Rebuild this as a component */
```
andlum, line 1: Can probs drop this comment since you did it! 😸

PR #112, opened 2022-03-04
remove more insights

nickpatrick:
continuation of #111


andlum (approved):
These changes look good. It'll take a day or 2 for them to be de-indexed in the search. Are there any other keywords we should look out for re: insights?

PR #122, opened 2022-04-01
Add docs for AWS S3

matthewisabel:
## What?
Add docs for AWS S3 integration beta

## Why?
New docs

## How?
New integration in beta

## Screenshot? (optional)
Screen Shot 2022-04-01 at 7 33 29 PM

## Anything Else? (optional)
This also reorders the integrations to be alphabetical except for webhooks which sits at the top.


docs/integrations/aws-s3.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+Radar can stream both events and locations to [AWS S3](https://aws.amazon.com/s3/).
+
+Use the AWS S3 integration to make Radar data available in a simple object storage format within your own systems, making your locations and events data more easily accessible by your engineers and data scientists.
+
+## Configuration
+
+While in beta, contact your customer success manager to enable this feature.
+
+The AWS S3 integration uses an [AWS Kinesis Data Firehose](https://aws.amazon.com/kinesis/data-firehose/) to stream new Radar locations and events. Additionally, [AWS KMS](https://aws.amazon.com/kms/) can be used [to encrypt Kinesis data streams server-side](https://docs.aws.amazon.com/streams/latest/dev/server-side-encryption.html).
```
jsani-radar, line 20: we should probably provide an example export file like we do for import, or explicitly call out the schema that'll land in S3

docs/integrations/aws-s3.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+<Alert alertType="info">
+  The AWS S3 integration is currently in <b>beta</b> and available on the {` `} 
+  <a href="https://radar.com/pricing" target="_blank">Enterprise plan</a>
+  .
+</Alert>
+
+Radar can stream both events and locations to [AWS S3](https://aws.amazon.com/s3/).
+
+Use the AWS S3 integration to make Radar data available in a simple object storage format within your own systems, making your locations and events data more easily accessible by your engineers and data scientists.
```
jsani-radar, line 14: mention some use cases here? some mentioning of data lakes (ex. Snowflake) or analytics (ex. Tableau) would be helpful to explain the rationale behind setting this up

sidebars.js
```diff
@@ -58,22 +58,23 @@ module.exports = {
       items: [
         { type: "doc", id: "integrations", label: "Overview" },
         "integrations/webhooks",
-        "integrations/braze",
         "integrations/urban-airship",
-        "integrations/onesignal",
-        "integrations/mparticle",
-        "integrations/segment",
+        "integrations/amplitude",
```
jsani-radar, line 62: nice, thanks for this

docs/integrations/aws-s3.mdx
matthewisabel: I'm not sure we need to at this point given it's documented as in beta. We could put the [base schema](https://github.com/radarlabs/infrastructure/blob/c59ac82af62c82e40529e738cddb2a78e3cc8e1e/utils/athena-schemas.ts#L3-L63), but I figured we could wait until the second customer given it can't be self-served anyway.

matthewisabel: I propose we add add once we're further in beta / exiting beta

jsani-radar (approved):

PR #130, opened 2022-04-13
Update to add remote tracking options

matthewisabel:
## What?
Add docs for remote tracking options

## Why?
New feature

## How?
New feature docs

### DIFF
--- a/docs/sdk/tracking.mdx
+++ b/docs/sdk/tracking.mdx
@@ -8,6 +8,7 @@ import TabItem from "@theme/TabItem";
 
 For background tracking, Radar's SDK supports custom tracking options as well as three presets: **`EFFICIENT`**, **`RESPONSIVE`**, and **`CONTINUOUS`**.
 
+#### Configuring Tracking Options
 While it's recommended to use one of the three optimized presets, custom parameters can also be passed at the time of tracking initialization. Additionally, you can also choose to forgo setting background tracking options as part of SDK initialization and override to any of the three preset on the Radar dashboard using [Remote Tracking Options](#remote-tracking-options).
 
 <Tabs
@@ -57,7 +58,8 @@ Radar.startTracking(trackingOptions)
   </TabItem>
 </Tabs>
 
-The SDK has two states: **stopped** and **moving**. `stopDuration` and `stopDistance` determine whether a location update is considered stopped, and the other options determine tracking behavior depending on state.
+#### Tracking states
+The SDK has two states: **stopped** and **moving**. When a track call is made, a user occupies one of these two states. `stopDuration` and `stopDistance` are Tracking Options that can be adjusted to alter when a user is assumed to be stopped, while other Tracking Options related to stopped and moving determine tracking behavior based on the user state.
 
 ## iOS
 
--- a/docs/sdk/tracking.mdx
+++ b/docs/sdk/tracking.mdx
@@ -3,9 +3,59 @@ sidebar_label: Tracking Options
 title: Tracking Options Reference
 ---
 
-For background tracking, the SDK supports custom tracking options as well as three presets: **`EFFICIENT`**, **`RESPONSIVE`**, and **`CONTINUOUS`**.
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
 
-Though we recommend using presets for most use cases, you can modify them.
+For background tracking, Radar's SDK supports custom tracking options as well as three presets: **`EFFICIENT`**, **`RESPONSIVE`**, and **`CONTINUOUS`**.
+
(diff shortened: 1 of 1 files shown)


docs/sdk/tracking.mdx
```diff
@@ -126,3 +176,11 @@ Option | Efficient | Responsive | Continuous
 **`syncGeofencesLimit`** | `10` | `10` | `0`
 **`foregroundService`** | `null` | `null` | `null`
 **`beacons`** | `false` | `false` | `false`
+
+## Remote tracking options
+
+SDK tracking options can be set [in dashboard settings](https://radar.com/dashboard/settings?#sdk-settings) for iOS and Android with the three default presets.
+
+<img src="https://images.ctfassets.net/f2vbu16fzuly/3fTIXlesjMjmtDLlS2yIIc/be9458fa994ae783242e8e5d01cfe697/tracking_options_ios.png" alt="Remote tracking options" />
```
jsani-radar, line 184: any chance we could get a higher quality image? it looks a bit grainy in the preview link

docs/sdk/tracking.mdx
```diff
@@ -3,9 +3,59 @@ sidebar_label: Tracking Options
 title: Tracking Options Reference
 ---
 
-For background tracking, the SDK supports custom tracking options as well as three presets: **`EFFICIENT`**, **`RESPONSIVE`**, and **`CONTINUOUS`**.
-
-Though we recommend using presets for most use cases, you can modify them.
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+For background tracking, Radar's SDK supports custom tracking options as well as three presets: **`EFFICIENT`**, **`RESPONSIVE`**, and **`CONTINUOUS`**.
```
jsani-radar, line 9: should we create a separate header for this section like "Initialize" or "Initialize background tracking"? feels like a bit of a grab bag of content

docs/sdk/tracking.mdx
```diff
@@ -3,9 +3,59 @@ sidebar_label: Tracking Options
...
+
+While it's recommended to use one of the three optimized presets, custom parameters can also be passed at the time of tracking initialization. Additionally, you can also choose to forgo setting background tracking options as part of SDK initialization and override to any of the three preset on the Radar dashboard using [Remote Tracking Options](#remote-tracking-options).
+
+<Tabs
```
jsani-radar, line 13: love the multi-language approach here, think we should keep propagating this throughout

jsani-radar (approved):
few small comments, not holding approval

lgtm and feel free to merge!

docs/sdk/tracking.mdx
```diff
@@ -3,11 +3,63 @@ sidebar_label: Tracking Options
...
-
-Though we recommend using presets for most use cases, you can modify them.
-
-The SDK has two states: **stopped** and **moving**. `stopDuration` and `stopDistance` determine whether a location update is considered stopped, and the other options determine tracking behavior depending on state.
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+For background tracking, Radar's SDK supports custom tracking options as well as three presets: **`EFFICIENT`**, **`RESPONSIVE`**, and **`CONTINUOUS`**.
+
+#### Configuring Tracking Options
```
nickpatrick, line 11: Nit: Sentence case?

docs/sdk/tracking.mdx
```diff
@@ -3,11 +3,63 @@ sidebar_label: Tracking Options
...
+```kotlin
+val trackingOptions = RadarTrackingOptions.RESPONSIVE
+trackingOptions.beacons = true
+Radar.startTracking(trackingOptions)
+```
+  </TabItem>
+</Tabs>
+
+#### Tracking states
+The SDK has two states: **stopped** and **moving**. When a track call is made, a user occupies one of these two states. `stopDuration` and `stopDistance` are Tracking Options that can be adjusted to alter when a user is assumed to be stopped, while other Tracking Options related to stopped and moving determine tracking behavior based on the user state.
```
nickpatrick, line 62: I could see us eventually expanding this into a more comprehensive "how the SDK works" section, but I think this is fine for now.

A few nits:
- `When a track call is made…` --> `After every location update…` (technically we recalculate stopped/moving state after every client-side location update, not after every track call)
- `can be adjusted to alter when a user is assumed to be stopped` (it should be extremely rare for users to need to change `stopDistance` and `stopDuration`, much more common to change the moving and stopped intervals, so we may want to note this or just omit this sentence entirely)

docs/sdk/tracking.mdx
matthewisabel: We seem to capitalize Tracking Options everywhere, same with Remote Tracking Options. It feels general so it's a bit odd. Didn't want to change it globally at the moment.

PR #133, opened 2022-04-15
Update trip index endpoint

kochis:
We no longer accept `active` as a parameter to the list trips endpoint. 

Instead, we use the `status` filter to retrieve trips with a given stats.

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1335,7 +1335,7 @@ Lists trips. Trips are sorted descending by `eta`.
 ###### Sample request
 
 ```bash
-curl "https://api.radar.io/v1/trips?active=true&destinationGeofenceTag=store&destinationGeofenceExternalId=123" \
+curl "https://api.radar.io/v1/trips?status=started&destinationGeofenceTag=store&destinationGeofenceExternalId=123" \
   -H "Authorization: prj_live_sk_..."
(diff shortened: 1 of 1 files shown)


docs/api.mdx
```diff
@@ -1324,7 +1324,7 @@ Lists trips. Trips are sorted descending by `eta`.
 
 ###### Query parameters
 
-- **`active`** (boolean, optional): If `true`, returns only active trips. If `false`, returns only stopped trips. If not specified, returns all trips.
+- **`status`** (string, optional): Retrieves trips by status. A string, comma-sparated including one or more of `pending`, `started`, `approaching`, `arrived`, `completed`, `canceled`, `expired`.
```
matthewisabel, line 1327:
```suggestion
- **`status`** (string, optional): Retrieves trips by status. A string, comma-separated including one or more of `pending`, `started`, `approaching`, `arrived`, `completed`, `canceled`, `expired`.
```

matthewisabel: @kochis should we add the behavior if no string is provided (all trips I'd guess)?

kochis: @matthewisabel Yeah, no status filter will return all trips

matthewisabel (approved):

PR #150, opened 2022-06-23
Place matching value props, advanced options

jsani-radar:
Per wg feedback, make the value props for place matching clearer, and add a section explaining the advanced options

### DIFF
--- a/docs/dashboard.mdx
+++ b/docs/dashboard.mdx
@@ -50,7 +50,7 @@ The User Activity page is the most straightforward way to debug a user's recent 
 
 ## Place matching
 
-Place matching is the simplest way to clean your geofence datasets while making sure they stay up-to-date and accurate over time. Navigate to the [geofences](https://radar.com/dashboard/geofences) page and select *Import*. If [places](/places) are enabled for your project, opt-in to this feature by toggling on *Enable place matching*. Select [chains](/places/chains) to match your geofences against. The import process will attempt to match each geofence against Radar's places for those chains.
+Place matching is the simplest way to make sure your geofences are accurate and stay up-to-date over time. Cleaning your geofence datasets via Radar's import process will reduce false positives by making sure that entry events are correct, trip approaching and arrived signals are improved, and messaging for nearby locations is relevant. Navigate to the [geofences](https://radar.com/dashboard/geofences) page and select *Import*. If [places](/places) are enabled for your project, opt-in to this feature by toggling on *Enable place matching*. Select [chains](/places/chains) to match your geofences against. The import process will attempt to match each geofence against Radar's places for those chains.
 
 ### How it works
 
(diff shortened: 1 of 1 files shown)


docs/dashboard.mdx
```diff
@@ -78,3 +78,9 @@ When an **isochrone** geofence is matched, the geofence will only adopt the cent
 When a **polygon** geofence is matched, the geofence will be linked to the Radar place, but not adopt the place center or geometry. This behavior supports geofences that represent areas such as parking lots and drive-thru lanes.
 
 ![Place matching polygon](/img/place-matching/place_matching_polygon.png)
+
+### Advanced options
+
+- **Search radius.** Place matching will look for the nearest location within 10 kilometers, consistent with our [search places](/api#search-places) endpoint. This value can be reduced if a tighter search window is desired.
+- **Geofence tag(s).** Filter the rows for which place matching is attempted. For example, place matching can only happen where the geofence tag is store, but not parking lot.
```
corypisano, line 85: maybe something like `For example, only perform place matching on rows where tag = "store" but not tag = "parking lot"`

docs/dashboard.mdx
```diff
@@ -50,7 +50,7 @@ The User Activity page is the most straightforward way to debug a user's recent
...
+Place matching is the simplest way to make sure your geofences are accurate and stay up-to-date over time. Cleaning your geofence datasets via Radar's import process will reduce false positives by making sure that entry events are correct, trip approaching and arrived signals are improved, and messaging for nearby locations is relevant. Navigate to the [geofences](https://radar.com/dashboard/geofences) page and select *Import*. If [places](/places) are enabled for your project, opt-in to this feature by toggling on *Enable place matching*. Select [chains](/places/chains) to match your geofences against. The import process will attempt to match each geofence against Radar's places for those chains.
```
corypisano, line 53: nit: maybe "accurate" instead of improved for "trip approaching and arrived signals are ..."?

corypisano (approved):
Lgtm, some small and/or ignorable copy nits

docs/dashboard.mdx
bschoeneweis, line 53: nit: also include exit events?

"will reduce false positives by making sure that entry and exit events are correct, trip approaching..."

bschoeneweis (approved):
Looks good.  Echoing @corypisano's comments as well but thanks for adding!

docs/dashboard.mdx
berylw, line 53: nit: maybe specify what we mean by accurate? "...make sure your geofence coordinates and geometries are accurate and stay up-to-date over time"

docs/dashboard.mdx
```diff
@@ -50,7 +50,7 @@ The User Activity page is the most straightforward way to debug a user's recent
 
 ## Place matching
```
berylw, line 52: wonder if we should do an upsell here with an info banner i.e. "Place matching requires that places are enabled for your project"

docs/dashboard.mdx
```diff
@@ -78,3 +78,9 @@ When an **isochrone** geofence is matched, the geofence will only adopt the cent
 When a **polygon** geofence is matched, the geofence will be linked to the Radar place, but not adopt the place center or geometry. This behavior supports geofences that represent areas such as parking lots and drive-thru lanes.
 
 ![Place matching polygon](/img/place-matching/place_matching_polygon.png)
+
+### Advanced options
+
+- **Search radius.** Place matching will look for the nearest location within 10 kilometers, consistent with our [search places](/api#search-places) endpoint. This value can be reduced if a tighter search window is desired.
```
berylw, line 84: nit maybe something like `By default, place matching will look for the nearest place within 10 kilometers (the max radius provided by our search places endpoint).`"

docs/dashboard.mdx
```diff
@@ -78,3 +78,9 @@ When an **isochrone** geofence is matched, the geofence will only adopt the cent
...
+- **Geofence tag(s).** Filter the rows for which place matching is attempted. For example, place matching can only happen where the geofence tag is store, but not parking lot.
+- **Use Radar's place geometry.** By default, if a place match is found, [circle geofences](/dashboard#circle) will adopt both the center and geometry of the Radar place. This behavior can be changed such that circle geofences only take on the center but not the polygon.
```
berylw, line 86: nit: should we use geometry to be consistent with the first sentence? `..circle geofences only take on the center but not the geometry`

docs/dashboard.mdx
jsani-radar: yeah, this sentence is a bit of a run-on sentence, will rework

PR #160, opened 2022-08-02
chore: [sc-11185] add fullstory to docs

corypisano:
Story details: https://app.shortcut.com/radarlabs/story/11185

More details on ticket

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -87,4 +87,10 @@ module.exports = {
       },
     ],
   ],
+  scripts: [
+    {
+      src: 'js/fullstory.js',
+      async: true,
+    }
+  ]
 };
--- a/src/theme/DocItem/index.js
+++ b/src/theme/DocItem/index.js
@@ -57,7 +57,7 @@
         }}
       />
 
-      <div className="row">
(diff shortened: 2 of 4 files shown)


corypisano:
k can see it gets unmasked halfway through here 
https://app.fullstory.com/ui/8TP1K/session/6012859036930048:(6012859036930048!5836067060518912(5010063723270144(1:1:936):6729735174819840(0)))

jsani-radar:
lgtm overall! thank you so, so much for this

do you have an example session in FS where it's functioning properly with the elements unmasked?

corypisano:
yea I added to the [ticket](https://app.shortcut.com/radarlabs/story/11185/add-fullstory-to-docs) comments

bschoeneweis:
Does anything here actually need to be masked?  Nothing really sensitive at all (or maybe I'm misunderstanding the use case of masking)

corypisano:
> Does anything here actually need to be masked? Nothing really sensitive at all (or maybe I'm misunderstanding the use case of masking)

Nope, added the unmasking at the top level. Just uses the same masking default as dashboard

jsani-radar (approved):

PR #162, opened 2022-08-03
typos and minor fixes from my onboarding diary

palmakkar:
## What?
Putting in minor fixes outlined in my onboarding doc:
- Typos
- Incorrect settings page referenced in toolkit guide - step is referencing app settings in toolkit and not radar settings
- disableAfter can be set for a time, which should be more relevant for real time delivery use cases

## Why?
Discovered some typos when onboarding

### DIFF
--- a/docs/toolkit.mdx
+++ b/docs/toolkit.mdx
@@ -40,7 +40,7 @@ To get started, you'll need a Radar account. Don't have a Radar account yet? [Si
 
 2. Open the app and grant notification permissions and location permissions.
 
-3. Copy your `Test Publishable` key from the *Keys* section of the [Settings page](https://radar.com/settings) and paste it into the app.
+3. Copy your `Test Publishable` key from the *Keys* section of the Settings page and paste it into the app.
 
 4. To track in the background, turn on *Tracking* (it is on by default). Then, select the *Tracking Options* preset to use, one of *Efficient*, *Responsive*, or *Continuous*. Learn about [iOS tracking presets](/sdk/tracking#ios-presets) and [Android tracking presets](/sdk/tracking#android-presets). Finally, on the notifications you want to receive, including *Events* (events received from Radar), *Locations* (location updates sent to Radar), and *Errors* (network, location, and permissions errors).
 
--- a/docs/toolkit.mdx
+++ b/docs/toolkit.mdx
@@ -40,14 +40,16 @@ To get started, you'll need a Radar account. Don't have a Radar account yet? [Si
 
 2. Open the app and grant notification permissions and location permissions.
 
-3. Copy your `Test Publishable` key from the *Keys* section of the Settings page and paste it into the app.
+3. Copy your `Test Publishable` key from the *Keys* section of the [Get Started page](https://radar.com/dashboard) and paste it into the app.
 
-4. To track in the background, turn on *Tracking* (it is on by default). Then, select the *Tracking Options* preset to use, one of *Efficient*, *Responsive*, or *Continuous*. Learn about [iOS tracking presets](/sdk/tracking#ios-presets) and [Android tracking presets](/sdk/tracking#android-presets). Finally, on the notifications you want to receive, including *Events* (events received from Radar), *Locations* (location updates sent to Radar), and *Errors* (network, location, and permissions errors).
+4. Open the *Settings* screen.
 
-5. Close the *Settings* screen.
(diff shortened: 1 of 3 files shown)


docs/toolkit.mdx
```diff
@@ -40,9 +40,9 @@ To get started, you'll need a Radar account. Don't have a Radar account yet? [Si
 
 2. Open the app and grant notification permissions and location permissions.
 
-3. Copy your `Test Publishable` key from the *Keys* section of the [Settings page](https://radar.com/settings) and paste it into the app.
+3. Copy your `Test Publishable` key from the *Keys* section of the Settings page and paste it into the app.
```
jsani-radar, line 43: fun fact the API keys can be found on the Settings page at the bottom
Screen Shot 2022-08-03 at 8 49 28 PM

I think instead we should say Get started and link out to `radar.com/dashboard` since that's the most prominent place they're located

docs/tutorials/building-a-delivery-tracker-with-live-location-and-etas.mdx
```diff
@@ -28,7 +28,7 @@ If you haven't already, sign up for Radar to get your API key. You can create up
 
 When a delivery is placed, call the [geofence upsert API](/api#upsert-a-geofence) to create a geofence for the delivery. In this case, use `delivery` for the geofence `tag` and the delivery ID for the geofence `externalId`.
 
-Optionally, set `disableAfter` to disable geofence entry events the delivery date, and set `userId` to the user ID of the delivery driver to enable geofence entry events only for that delivery driver.
+Optionally, set `disableAfter` to disable geofence entry events the delivery time, and set `userId` to the user ID of the delivery driver to enable geofence entry events only for that delivery driver.
```
jsani-radar, line 31: should this be "disable geofence entry events _after_ the delivery time"?

jsani-radar (commented):
few small nits, otherwise lgtm!

will be 👍🏽 on next pass

docs/toolkit.mdx
palmakkar: Ah yes that makes sense. 

I think the bigger issue I had which I hadn't articulated well yet is that step 4 is referring to actions you should take on the settings page of the _toolkit app_ and not the dashboard settings. And step 5 also says to close the settings screen. There's no mention of opening the settings screen though so I was wondering if in this step we're conflating the settings instruction in step 3 with the settings for the app.  Maybe we add another step to open the settings screen in the toolkit app?

docs/tutorials/building-a-delivery-tracker-with-live-location-and-etas.mdx
palmakkar: Nice catch, will fix

docs/toolkit.mdx
jsani-radar: yup, that's a great call - let's add a new step 4 that mentions opening the settings screen in toolkit to clarify and be consistent with the separate step for closing

docs/tutorials/building-a-delivery-tracker-with-live-location-and-etas.mdx
jsani-radar: ^ friendly bump here

docs/toolkit.mdx
```diff
@@ -40,14 +40,16 @@ To get started, you'll need a Radar account. Don't have a Radar account yet? [Si
...
+7. To track once in the foreground, tap the refresh button! **If you correctly granted permissions and pasted your API key, and if you have connectivity, you will see your current state in the app and see your user on the [Users page](https://radar.com/users) in the dashboard.** If you are in a geofence or at a place, you will see events in the app and see events on the [Events page](https://radar.com/events) in the dashboard.
+
+8. To track in the background, move more than 100 meters! Depending on the *Tracking Options* preset you selected, your location will update in the background. **Note that location updates may be delayed significantly by Android [Doze Mode and App Standby](https://developer.android.com/training/monitoring-device-state/doze-standby.html), iOS [Low Power Mode](https://support.apple.com/en-us/HT205234), or if the device has connectivity issues, low battery, or wi-fi disabled.**
```
jsani-radar, line 53: know this was just lifted / shifted, but it's interesting that this step and #5 start with the same "To track in the background" - wonder if we want to eventually collapse / colocate those

jsani-radar (approved):
small nit for a potential follow-up PR

otherwise lgtm - send it 🚀

PR #163, opened 2022-08-04
Add geofence.minimumConfidence to geofence schema

ahwreck:
Include new `geofence.minimalConfidence` parameter to upsert geofence API docs


jsani-radar:
good call adding this, but I don't think we support updating this field in the upsert endpoint today

I'll move it over to the Geofence schema

jsani-radar:
we actually don't expose this field (yet) in our geofence update paths - working on how we want to move forward with making this self-serve, even if it's just SU for now

jsani-radar:
closing as the ticket to expose this functionality outside of SU has been backlogged - https://linear.app/radarlabs/issue/CORE-13/support-geofenceminimumconfidence-in-geofence-create-update-paths


will reopen / make changes when that's shipped

PR #165, opened 2022-08-10
Add place.chain.metadata to the SFMC integration

bschoeneweis:
## What?
Add place.chain.metadata to the SFMC integration.


jsani-radar (approved):

jsani-radar:
let's merge once the server pieces are shipped

PR #167, opened 2022-08-15
regenerate chains docs and update schema

brettguenther:
## What?

Update chains list to latest

## Why?

We have added new chains


jsani-radar (approved):

jsani-radar:
we should prob have a bot / cron doing this ~weekly or so, similar to how we refresh our server data

PR #168, opened 2022-08-15
Document debug logging listeners

jsani-radar:
follow-up from @palmakkar's onboarding that we don't document the debug log message listeners

### DIFF
--- a/docs/sdk/ios.mdx
+++ b/docs/sdk/ios.mdx
@@ -874,7 +874,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, RadarDelegate {
     // do something with status
 }
 
-- (void)didLog:(NSString *_Nonnull)message {
+- (void)didLogMessage:(NSString *_Nonnull)message {
     // print message for debug logs
 }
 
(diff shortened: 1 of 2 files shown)


jsani-radar:
Screen Shot 2022-08-15 at 4 43 25 PM

Screen Shot 2022-08-15 at 4 43 48 PM

docs/sdk/ios.mdx
```diff
@@ -870,6 +874,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, RadarDelegate {
     // do something with status
 }
 
+- (void)didLog:(NSString *)message {
```
jrtibbetts, line 877: `- (void)didLogMessage:(NSString *_Nonnull)message {`

jrtibbetts (changes requested):
Just one small change for the Objective-C example.

docs/sdk/ios.mdx
jsani-radar: hmm it doesn't seem like we add this annotation for any of the other listeners - what's the benefit / drawback of including / not including it

jrtibbetts: It's fine to leave out the `_Nonnull`, but the user may get a compiler warning for it. But the signature _does_ have to be `didLogMessage`, not just `didLog`.

jrtibbetts (approved):

PR #169, opened 2022-08-17
Update metadata fields for SFMC integration

bschoeneweis:
## What?
Updating the SFMC integration event field types and names.

### DIFF
--- a/docs/integrations/salesforce.mdx
+++ b/docs/integrations/salesforce.mdx
@@ -42,7 +42,7 @@ Radar Event Field | Salesforce Event Attribute | Type | Example
 `geofence.description` | `geofenceDescription` | string | `"Store #123"`
 `geofence.tag` | `geofenceTag` | string | `"store"`
 `geofence.externalId` | `geofenceExternalId` | string | `"123"`
-`geofence.metadata[{key}]` | `geofenceMetadata{key}` | {type} | `{value}`
+`geofence.metadata` | `geofenceMetadata` | string | `'{"parking":false}'`
 `confidence` | `confidence` | string | `"high"`
 
 If [Regions](/regions) is enabled, Radar will also send the following attributes:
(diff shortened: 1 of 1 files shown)


docs/integrations/salesforce.mdx
```diff
@@ -42,7 +42,7 @@ Radar Event Field | Salesforce Event Attribute | Type | Example
 `geofence.description` | `geofenceDescription` | string | `"Store #123"`
 `geofence.tag` | `geofenceTag` | string | `"store"`
 `geofence.externalId` | `geofenceExternalId` | string | `"123"`
-`geofence.metadata[{key}]` | `geofenceMetadata{key}` | {type} | `{value}`
+`geofence.metadata` | `geofenceMetadata` | string | `'{"parking":false}'`
```
jsani-radar, line 45: interesting, this is a new pattern across our integrations, as others loop through `metadata` and create keys like `geofence_metadata_{key}` (https://radar.com/documentation/integrations/braze)

do you think this new approach is better suited for SFMC because of how strict the schema is?

jsani-radar (commented):
looking solid overall - just want to make sure we're aligned on the schema approach for SFMC going forward

docs/integrations/salesforce.mdx
bschoeneweis: Yeah this is sort of the only way to pass dynamic metadata to SFMC and have it be accepted since the schema needs to be decided in advance.  Another option could be integration settings and defining key/values (and types), but that seems a bit cumbersome

jsani-radar: nah don't want to go too far down that rabbit hole - we have the pattern for Olo in the dashboard, but dynamically updating the downstream event definition / key doesn't seem ideal

bschoeneweis: Right and yeah you can't actually update the event definition w/ the SFMC API, you have to create a new event definition entirely and that would cause some breaking changes/unexpected behavior

jsani-radar: lol yeah let's absolutely not do that

jsani-radar (approved):

PR #173, opened 2022-08-24
Add approachingThreshold to trip schema and PATCH call

bschoeneweis:
## What?
Adding `approachingThreshold` to the trip schema and `PATCH` call.

### Trips schema
image

### `PATCH` call
image


jsani-radar (approved):
lgtm

palmakkar:
nit: The units may fit better in the fit in the first sentence without parenthesis - "the approaching threshold for the trip in minutes" to have it a bit more consistent with the sentence for eta, and then the second sentence just clarifies the override

PR #175, opened 2022-08-24
Move the units to the first sentence for trip approaching thresholds

bschoeneweis:
## What?
Small follow-on tweak for formatting.


jsani-radar (approved):

jsani-radar:
@palmakkar feel free to start approving and merging these for your workstream

not sure if there's a GH setting we need to enable for that

PR #190, opened 2022-10-14
Remove constraints on size to match product

matthewisabel:
Update for polygon changes


nickpatrick:
Do we need the note with the sentence above it?

nickpatrick:
Sorry, misread the diff on my phone, lgtm

Should we do an info bubble instead of text in italics?

matthewisabel:
Just moved it below. Hadn't seen that other text.

jsani-radar (approved):

PR #213, opened 2022-12-07
capacitor docs update for trips revamp changes

palmakkar:
## What?

capacitor docs update for trips revamp changes

### DIFF
--- a/docs/sdk/capacitor.mdx
+++ b/docs/sdk/capacitor.mdx
@@ -361,7 +361,7 @@ Radar.completeTrip().then((result) => {
 });
 ```
 
-Or, to cancel the trip and stop tracking, call:
+Or, to cancel the trip and revert tracking, call:
 
 ```javascript
 Radar.cancelTrip().then((result) => {
--- a/docs/sdk/capacitor.mdx
+++ b/docs/sdk/capacitor.mdx
@@ -377,7 +377,7 @@ Radar.startTrip({
 });
 ```
 
-Later, to complete the trip and revert tracking, call:
+Later, complete the trip by calling:
 
 ```javascript
 Radar.completeTrip().then((result) => {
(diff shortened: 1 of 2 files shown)


docs/sdk/capacitor.mdx
```diff
@@ -305,29 +305,27 @@ Radar.startTrip({
     destinationGeofenceTag: 'store',
     destinationGeofenceExternalId: '123',
     mode: 'car'
-  }
+  },
+  trackingOptions: RadarTrackingOptions.presetContinuous
```
tjulien, line 309: I don't think we expose these presets in Capacitor (possibly not in the other x-platform as well)

docs/sdk/capacitor.mdx
```diff
@@ -346,29 +350,47 @@ Radar.startTrip({
...
+    "foregroundServiceEnabled": false
   }
 }).then((result) => {
   // do something with result.status
 });
-Radar.startTrackingContinuous();
 ```
 
-Later, to complete the trip and stop tracking, call:
+Later, to complete the trip and revert tracking, call:
```
jsani-radar, line 380: just having "revert" here is a bit confusing

in the native SDKs you added - "If tracking was disabled before the trip started, it will stop after the trip ends. Otherwise, it will revert to the tracking options in use before the trip started"

I think this better reads - "Later, complete the trip by calling:" and then having a reminder below the code block about the above

docs/sdk/capacitor.mdx
```diff
@@ -346,29 +350,47 @@ Radar.startTrip({
...
 
 ```javascript
 Radar.completeTrip().then((result) => {
   // do something with result.status
 });
-Radar.stopTracking();
 ```
 
-Or, to cancel the trip and stop tracking, call:
+Or, to cancel the trip and revert tracking, call:
```
jsani-radar, line 388: ditto

jsani-radar (commented):
one small comment in two spots

docs/sdk/capacitor.mdx
```diff
@@ -346,29 +350,47 @@ Radar.startTrip({
     destinationGeofenceTag: 'store',
     destinationGeofenceExternalId: '123',
     mode: 'car'
+  },
```
tjulien, line 353: the new params format is:
```
Radar.startTrip({
  options: {
    tripOptions { ... },
    trackingOptions: { ... },
  }
})
```

docs/sdk/capacitor.mdx
palmakkar: agree that's more clear, updated

brettguenther (approved):

PR #214, opened 2022-12-07
switch to using gtag instead of UA for google analytics

kochis:


jsani-radar:
closing out for now, Linear ticket to track: https://linear.app/radarlabs/issue/CORE-364/upgrade-docusaurus-to-a-stable-version

PR #220, opened 2022-12-16
Adding custom event example to React Native docs

ahwreck:
Adding example for new custom event functionality

### DIFF
--- a/docs/sdk/react-native.mdx
+++ b/docs/sdk/react-native.mdx
@@ -543,7 +543,20 @@ Radar.getMatrix({
   // optionally, do something with err
 });
 ```
+#### Custom events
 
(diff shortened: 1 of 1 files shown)


brettguenther:
@jsani-radar - do we want to change the header structure while we are at it? A. a bit sparse for the amount of content B. nesting everything one level under integrate (i.e. battery usage as a header)

Screen Shot 2023-01-13 at 8 42 11 AM

jsani-radar:
hmm, yeah agree that the layout is strange - interesting how our x-platform sidebars all look like that whereas the native ones look like the below
Screenshot 2023-01-21 at 12 23 11 AM

thinking it makes sense to remove the Integrate header entirely, then Other APIs becomes top-level so it can have nested subsections

jsani-radar (approved):

PR #221, opened 2022-12-18
Remove vercel noindex header

kochis:
Now that we're hosting the docs on Vercel, we need to make sure this header is not included in the response.

Preview deployments should already include this header by default: https://vercel.com/guides/are-vercel-preview-deployment-indexed-by-search-engines

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -5,11 +5,5 @@
   "redirects": [
     { "source": "/", "destination": "/documentation/", "permanent": true }
   ],
-  "headers": [
-    {
-      "source": "/(.*)",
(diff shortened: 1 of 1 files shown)


kochis:
Confirmed that it's still present on preview environment

image

kochis:
Confirmed fix on production

image

jsani-radar:
good catch - how'd you discover this?

also, did this mean that our docs were never getting indexed?

kochis:
> also, did this mean that our docs were never getting indexed?

Just since we switched over to using the vercel docs, which was about a week ago. We got a search console warning about the noindex header.

PR #223, opened 2023-01-03
Instruction for how to generate Olo API Key

JoRoRadar:
To generate the keys needed for the integration you need to reach out to the Olo Support Agent to have them generated for you. This isn't called out in our documentation and is causing confusion.


brettguenther:
@ahwreck - this is changing soon, correct? if so, we should just wait until we are cutover to the new path.

jsani-radar (approved):

PR #229, opened 2023-01-24
Fix broken links on presets page

lmeier:


jsani-radar (approved):
good catches! thanks for fixing these

we've had to do this a few times as Apple annoyingly keeps changing the URLs in their docs 🙄

PR #235, opened 2023-01-27
update api autocomplete and fwd geocode responses

brettguenther:
## What?

Update api responses for forward geocode and autocomplete

## Why?

Attributes out of date and autocomplete place no longer existings


jsani-radar:
squinting at the diff I'm not seeing which attributes are out of date - seems like the fields are the same but just in a different order?

approving and merging anyways

jsani-radar (approved):

PR #242, opened 2023-02-09
delay detection docs

palmakkar:
## What?

Adding documentation for the new delay field on trips
## Why?

Delay detection launch

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1291,8 +1291,8 @@ A trip is a sequence of location updates with metadata and a unique ID, optional
     }
   ],
   "delay": {
-      "delayed": false,
-      "scheduledArrivalTimeDelay": 0
+    "delayed": false,
+    "scheduledArrivalTimeDelay": 0
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


docs/trip-tracking.mdx
```diff
@@ -368,6 +368,12 @@ You can receive events client-side via the [SDK](/sdk) or server-side via [event
 }
 ```
 
+## Delay detection
+Delay detection flags trips at risk of being delayed or incomplete. To enable this feature, configure a *Delay Threshold* on the  [Project settings](https://radar.com/dashboard/settings#trips-settings) page under *Trips*. Then, ensure that `scheduledArrivalAt` is set for trips.  When a trip is predicted to be delayed by more than the chosen threshold, it will be marked as delayed.
```
corypisano, line 372: just a nit on copy, had to reread "Delay detection flags trips .." phrase. Think my mind initially read `flags` as a noun, like we set "Delay detection flags"

maybe something like:
"Delay detection identifies trips that are at risk"

docs/api.mdx
```diff
@@ -1288,7 +1289,11 @@ A trip is a sequence of location updates with metadata and a unique ID, optional
       ],
       "updatedAt": "2020-08-20T10:30:55.837Z"
     }
-  ]
+  ],
+  "delay": {
+      "delayed": false,
+      "scheduledArrivalTimeDelay": 0
```
corypisano, line 1295: nit: indentation should be 2 spaces for consistency with above

docs/trip-tracking.mdx
palmakkar: yea fair enough

jsani-radar (approved):

PR #244, opened 2023-02-22
Add note for capacitor listeners

corypisano:

### DIFF
--- a/docs/sdk/capacitor.mdx
+++ b/docs/sdk/capacitor.mdx
@@ -313,7 +313,7 @@ Radar.addListener('log', (result) => {
 });
 ```
 
-Add event listeners outside of your view lifecycle if you want them to work when the app is in the background.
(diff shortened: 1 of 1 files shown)


docs/sdk/capacitor.mdx
```diff
@@ -313,7 +313,7 @@ Radar.addListener('log', (result) => {
 });
 ```
 
-Add event listeners outside of your view lifecycle if you want them to work when the app is in the background.
+Add event listeners outside of your view lifecycle if you want them to work when the app is in the background. If you are using a foreground service on Android, it is possible for it to outlive the javascript engine due to Capacitor destroying the WebView when the activity is destroyed (e.g. swiped away). Logic that needs to outlive the WebView should be performed server side.
```
corypisano, line 316: word smithing suggestions welcome

brettguenther: Should we be more forward here on the recommended path? Basically something to the effect of:

```
Capacitor event listener firing will only function while the Capacitor webview and associated javascript engine are alive. These services are shutdown when the associated Android activity is killed. 

If performing background location processing, it is recommended to implement native event emitters (link to android receiver and iOS delegate docs), or leverage Radar's server side integrations (link to integrations).
```

Is that a correct interpretation? Is there anything we can reference on iOS behavior?

jsani-radar (approved):

PR #250, opened 2023-02-28
Add offline replay to tracking options docs

jsani-radar:
https://linear.app/radarlabs/issue/CORE-307/add-offline-replay-tracking-option-to-docs


lmeier (approved):
👍

PR #254, opened 2023-03-03
Add new canadian provinces

j949wang:
## What?

Add new Canadian provinces to documentation

## Why?

With the introduction of the international regions v1 pipeline. We are starting to introduce new regions.

## Screenshots (optional)

Screen Shot 2023-03-03 at 15 41 02
Screen Shot 2023-03-03 at 15 40 51

### DIFF
--- a/docs/regions.mdx
+++ b/docs/regions.mdx
@@ -59,7 +59,7 @@ Regions can also be used to allowlist or blocklist location updates in specific 
 
 ## States
 
-Regions supports state detection in the US only. Each state has a `name` and a unique 2-letter `code`. View the [full list of states](/regions/states).
+Regions supports state detection in the US and Canada only. Each state has a `name`, a unique 2-letter `code`, and the residing `country`. View the [full list of states](/regions/states).
 
 ## DMAs
 
--- a/docs/regions.mdx
+++ b/docs/regions.mdx
@@ -59,7 +59,7 @@ Regions can also be used to allowlist or blocklist location updates in specific 
 
 ## States
 
-Regions supports state detection in the US and Canada only. Each state has a `name`, a unique 2-letter `code`, and the residing `country`. View the [full list of states](/regions/states).
+Regions supports state detection in the US and Canada only. Each state has a `name` and a 2-letter `code` unique to the residing `country`. View the [full list of states](/regions/states).
 
 ## DMAs
 
--- a/docs/regions/states.mdx
+++ b/docs/regions/states.mdx
@@ -6,73 +6,73 @@ title: States
 
 ## United States
 
-Code | Country | Name  
+Code | Name | Country
 --- |  --- | ---
-`AK` | `United States` | `Alaska`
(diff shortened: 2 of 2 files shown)


docs/regions/states.mdx
```diff
@@ -2,60 +2,73 @@
...
-`WA` | `Washington`
-`WI` | `Wisconsin`
-`WV` | `West Virginia`
-`WY` | `Wyoming`
+Code | Country | Name  
+--- |  --- | ---
+`AK` | `United States` | `Alaska`
+`AL` | `United States` | `Alabama`
+`AR` | `United States` | `Arkansas`
+`AB` | `Canada` | `Alberta`
```
david-goodfellow, line 14: Up to product but thinking we may want to group them by country. Like put all United States then all canada

docs/regions/states.mdx
```diff
@@ -2,60 +2,73 @@
 title: States
 ---
 
-[Regions](/regions) supports state detection in the US only. Each state has a `name` and a unique 2-letter `code`.
+[Regions](/regions) supports state detection in the US and Canada only. Each state has a `name`, a unique 2-letter `code` and the residing `country`.
 
 ## State list
```
jsani-radar, line 7: not a big fan of "State list" as the top-level header and only href on the page

let's remove this and split it out so we have:
## United states
[table with US states]

## Canada
[table with Canadian provinces]

jsani-radar (commented):
small comment about formatting, content lgtm

will be 👍🏽 on next pass

docs/regions/states.mdx
```diff
@@ -2,60 +2,73 @@
 title: States
 ---
 
-[Regions](/regions) supports state detection in the US only. Each state has a `name` and a unique 2-letter `code`.
+[Regions](/regions) supports state detection in the US and Canada only. Each state has a `name`, a unique 2-letter `code` and the residing `country`.
```
jsani-radar, line 5: nit: comma usage should be consistent with top-level Regions page

docs/regions.mdx
```diff
@@ -59,7 +59,7 @@ Regions can also be used to allowlist or blocklist location updates in specific
 
 ## States
 
-Regions supports state detection in the US only. Each state has a `name` and a unique 2-letter `code`. View the [full list of states](/regions/states).
+Regions supports state detection in the US and Canada only. Each state has a `name`, a unique 2-letter `code`, and the residing `country`. View the [full list of states](/regions/states).
```
jsani-radar, line 62: nit: the uniqueness is on the 2-letter code AND country, correct?

docs/regions/states.mdx
j949wang: Sounds good, should we still keep the `country` column?

jsani-radar: country column is fine, but let's make that last so it's easy to see the correspondence between the 2-letter code and name

jsani-radar (approved):

PR #255, opened 2023-03-08
recommend a specific version

tjulien:
So that users control upgrading by default.


docs/sdk/web.mdx
```diff
@@ -15,7 +15,7 @@ See the source on GitHub [here](https://github.com/radarlabs/radar-sdk-js). Or,
 In an HTML page, include the SDK using a `<script>` tag:
 
 ```html
-<script src="https://js.radar.com/v3/radar.min.js"></script>
+<script src="https://js.radar.com/v3.6.1/radar.js"></script>
```
kochis, line 18: We should still recommend the `radar.min.js` version here.

kochis (approved):

PR #257, opened 2023-03-10
Add address validation endpoint and expandUnits option in autocomplete

palmakkar:
## What?

add documentation for address validation and the expandUnits option in autocomplete

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -580,7 +580,7 @@ Autocompletes partial addresses and place names, sorted by relevance.
 - **`layers`** (string, optional): Optional layer filters. A string, comma-separated, including one or more of `place`, `address`, `intersection`, `street`, `neighborhood`, `postalCode`, `locality`, `county`, `state`, `country`, `coarse`, and `fine`. Note that `coarse` includes all of `neighborhood`, `postalCode`, `locality`, `county`, `state`, and `country`, whereas `fine` includes all of `place`, `address`, `intersection`, and `street`. If not provided, results from all layers will be returned.
 - **`limit`** (number, optional): The max number of addresses to return. A number between 1 and 100. Defaults to 10.
 - **`country`** (string, optional): An optional countries filter. A string of comma separated countries, the unique 2-letter [country code](/regions/countries).
-- **`expandUnits`** (boolean, optional): If `true`, separate results will be returned for each unit in a building. Only available for the `address` layer and in the countries `US` and `CA`. Defaults to `false`.
+- **`expandUnits`** (boolean, optional): If `true`, separate results will be returned for each unit in a building. Only available when `layers="address"` for the countries `US` and `CA`. Defaults to `false`.
 
 ###### Authentication level
 
@@ -625,7 +625,6 @@ curl "https://api.radar.io/v1/search/autocomplete?query=brooklyn+roasting&near=4
       "stateCode": "NY",
       "state": "New York",
       "street": "Clinton St",
-      "unit": "Apt 3"
       "layer": "place",
       "formattedAddress": "1 Clinton St, Brooklyn, New York, NY 11201 USA",
       "placeLabel": "Brooklyn Roasting Company"
@@ -813,35 +812,34 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
 }
 ```
 
-#### Validate address
+#### Validate an address
 
 Validates an address. Currently only available for US and Canada addresses.
 
 This endpoint is best for validating complete structured addresses. For partial addresses or place names, call the [autocomplete](#autocomplete) endpoint instead.
 
-Queries must include a either a combination of `number`, `street`, and `unit` if applicable, or an `addressLabel`.
-
 Returns a `verificationStatus` as part of the `result`, as defined below:
 
-- **`Verified`**: The input has a complete and unique match with a result.
-- **`Partially Verified`**: The input has a partial match with a result.
-- **`Ambiguous`**: The input closely matches multiple results.
-- **`Unverified`**: Unable to verify the input address.
+- **`verified`**: The input has a complete and unique match with a result.
+- **`partially verified`**: The input has a partial match with a result.
+- **`ambiguous`**: The input closely matches multiple results.
+- **`unverified`**: Unable to verify the input address.
 
 ###### Definitions
 
-<span className="badge badge--success">GET</span> <code>https://api.radar.io/v1/address/validate</code>
+<span className="badge badge--success">GET</span> <code>https://api.radar.io/v1/addresses/validate</code>
 
 ###### Query parameters
 
-- **`number`** (string, optional): The building or P.O. box number.
-- **`street`** (string, optional): The street name.
-- **`unit`** (string, optional): The unit number, such as apartment or suite (e.g. Apt 3).
-- **`addressLabel`** (string, optional): The first line of the address, including the building number, street name, and unit number if applicable (e.g. 20 Jay St Apt 3).
 - **`city`** (string, required): The city name.
-- **`stateCode`** (string, required): The state abbreviation (e.g. NY).
+- **`stateCode`** (string, required): The 2-letter state abbreviation (e.g., `NY`).
 - **`postalCode`** (string, required): The postal or zip code. Plus 4 codes are accepted but not required.
-- **`countryCode`** (string, required): The country abbreviation (e.g. US).
+- **`countryCode`** (string, required): The unique 2-letter [country code](/regions/countries) (e.g., `US`).
+- **`number`** (string, optional): The building or P.O. box number. If not provided, `addressLabel` must be provided. 
+- **`street`** (string, optional): The street name. If not provided, `addressLabel` must be provided.
+- **`unit`** (string, optional): The unit number, such as apartment or suite (e.g., `Apt 3`).
+- **`addressLabel`** (string, optional): The first line of the address, including the building number, street name, and unit number if applicable (e.g., `20 Jay St Apt 3`). If not provided, `number` and `street` must be provided.
+
 
 ###### Authentication level
 
@@ -854,7 +852,7 @@ Publishable
 ###### Sample request
 
 ```bash
-curl "https://api.radar.io/v1/addresses/validate?qountryCode=US&stateCode=NY&city=New%20York&number=20&postalCode=11201&street=Jay%20St&unit=Apt%203" \
+curl "https://api.radar.io/v1/addresses/validate?countryCode=US&stateCode=NY&city=New%20York&number=841&postalCode=10003&street=Broadway%20St&unit=Fl%7" \
   -H "Authorization: prj_live_pk_..."
 ```
 
@@ -866,32 +864,24 @@ curl "https://api.radar.io/v1/addresses/validate?qountryCode=US&stateCode=NY&cit
     "code": 200
   },
   "address": {
-    "latitude": 40.7041,
-    "longitude": -73.9867,
-    "geometry": {
-      "type": "Point",
-      "coordinates": [-73.9867,40.7041]
-    },
     "country": "United States",
     "countryCode": "US",
-    "county": "Kings County",
-    "city": "Brooklyn",
-    "number": "20",
-    "neighborhood": "DUMBO",
-    "postalCode": "11201",
-    "plus4": "8301"
+    "city": "New York",
+    "number": "841",
+    "postalCode": "10003",
+    "plus4": "4704"
     "stateCode": "NY",
     "state": "New York",
-    "street": "Jay St",
-    "unit": "Apt 3",
-    "formattedAddress": "20 Jay St, Brooklyn, New York, NY 11201 USA",
+    "street": "Broadway",
+    "unit": "Fl 7",
+    "formattedAddress": "841 Broadway Fl 7, New York, NY 10003-4704, USA",
     "addressLabel": "20 Jay St",
     "metadata": {
       "propertyType": "residential"
     }
   },
   "result": {
-    "verificationStatus": "Verified"
+    "verificationStatus": "verified"
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/api.mdx
```diff
@@ -624,6 +625,7 @@ curl "https://api.radar.io/v1/search/autocomplete?query=brooklyn+roasting&near=4
       "stateCode": "NY",
       "state": "New York",
       "street": "Clinton St",
+      "unit": "Apt 3"
```
bschoeneweis, line 628: May want a separate example for using `expandUnits` that include the new response fields.  Don't think this query would return `unit`

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+This endpoint is best for validating complete structured addresses. For partial addresses or place names, call the [autocomplete](#autocomplete) endpoint instead.
+
+Queries must include a either a combination of `number`, `street`, and `unit` if applicable, or an `addressLabel`.
+
+Returns a `verificationStatus` as part of the `result`, as defined below:
+
+- **`Verified`**: The input has a complete and unique match with a result.
+- **`Partially Verified`**: The input has a partial match with a result.
+- **`Ambiguous`**: The input closely matches multiple results.
+- **`Unverified`**: Unable to verify the input address.
```
bschoeneweis, line 829: Probably want these lowercase since people will be string matching

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+
+###### Definitions
+
+<span className="badge badge--success">GET</span> <code>https://api.radar.io/v1/address/validate</code>
```
bschoeneweis, line 833: `/addresses/validate`

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+###### Query parameters
+
+- **`number`** (string, optional): The building or P.O. box number.
+- **`street`** (string, optional): The street name.
+- **`unit`** (string, optional): The unit number, such as apartment or suite (e.g. Apt 3).
+- **`addressLabel`** (string, optional): The first line of the address, including the building number, street name, and unit number if applicable (e.g. 20 Jay St Apt 3).
+- **`city`** (string, required): The city name.
+- **`stateCode`** (string, required): The state abbreviation (e.g. NY).
+- **`postalCode`** (string, required): The postal or zip code. Plus 4 codes are accepted but not required.
+- **`countryCode`** (string, required): The country abbreviation (e.g. US).
```
bschoeneweis, line 844: Also think we may want to flip the order on these so the required fields are up top (but I also get laying it out like an address).  Will leave to you to consider

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+    "street": "Jay St",
+    "unit": "Apt 3",
+    "formattedAddress": "20 Jay St, Brooklyn, New York, NY 11201 USA",
+    "addressLabel": "20 Jay St",
+    "metadata": {
+      "propertyType": "residential"
+    }
+  },
+  "result": {
+    "verificationStatus": "Verified"
```
bschoeneweis, line 894: nit: lowercase for `verified`

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+    "city": "Brooklyn",
+    "number": "20",
+    "neighborhood": "DUMBO",
+    "postalCode": "11201",
+    "plus4": "8301"
+    "stateCode": "NY",
+    "state": "New York",
+    "street": "Jay St",
+    "unit": "Apt 3",
+    "formattedAddress": "20 Jay St, Brooklyn, New York, NY 11201 USA",
```
bschoeneweis, line 887: nit: `formattedAddress` should have the `unit` and `plus4`

`"formattedAddress": "20 Jay St Apt 3, Brooklyn, New York, NY 11201-8301 USA",`

docs/api.mdx
```diff
@@ -580,6 +580,7 @@ Autocompletes partial addresses and place names, sorted by relevance.
 - **`layers`** (string, optional): Optional layer filters. A string, comma-separated, including one or more of `place`, `address`, `intersection`, `street`, `neighborhood`, `postalCode`, `locality`, `county`, `state`, `country`, `coarse`, and `fine`. Note that `coarse` includes all of `neighborhood`, `postalCode`, `locality`, `county`, `state`, and `country`, whereas `fine` includes all of `place`, `address`, `intersection`, and `street`. If not provided, results from all layers will be returned.
 - **`limit`** (number, optional): The max number of addresses to return. A number between 1 and 100. Defaults to 10.
 - **`country`** (string, optional): An optional countries filter. A string of comma separated countries, the unique 2-letter [country code](/regions/countries).
+- **`expandUnits`** (boolean, optional): If `true`, separate results will be returned for each unit in a building. Only available for the `address` layer and in the countries `US` and `CA`. Defaults to `false`.
```
bschoeneweis, line 583: nit: might even be more specific and say when `layers=address` and ... so it's clear it can only be the address layer

docs/api.mdx
jsani-radar: nit: trailing `,`

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
 }
 ```
 
+#### Validate address
```
jsani-radar, line 816: nit: "Validate an address" as our other APIs read "Get a geofence" and "Create a new trip"

docs/api.mdx
jsani-radar: agreed, let's follow the same pattern we do for `confidence` in forward geocoding https://radar.com/documentation/api#forward-geocode

docs/api.mdx
jsani-radar: is this meant to operate on one address or many addresses?

docs/api.mdx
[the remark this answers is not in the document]
jsani-radar: agreed, let's follow the pattern we have in autocomplete https://radar.com/documentation/api#query-parameters-3

jsani-radar: +1, don't think it's problematic / unintuitive if we go city, state, country, postal, number, stress, unit, address label

docs/api.mdx
jsani-radar: for apt numbers, do they need a `#` prefix?

docs/api.mdx
jsani-radar: curious why we decided to tuck this in a `result` key rather than having it within `address`?

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+
+Validates an address. Currently only available for US and Canada addresses.
+
+This endpoint is best for validating complete structured addresses. For partial addresses or place names, call the [autocomplete](#autocomplete) endpoint instead.
+
+Queries must include a either a combination of `number`, `street`, and `unit` if applicable, or an `addressLabel`.
```
jsani-radar, line 822: let's follow what we do with search places today in terms of describing a combo of params required https://radar.com/documentation/api#query-parameters-6

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+
+###### Definitions
+
+<span className="badge badge--success">GET</span> <code>https://api.radar.io/v1/address/validate</code>
+
+###### Query parameters
+
+- **`number`** (string, optional): The building or P.O. box number.
+- **`street`** (string, optional): The street name.
+- **`unit`** (string, optional): The unit number, such as apartment or suite (e.g. Apt 3).
```
jsani-radar, line 839: nit: `e.g.` should be followed by a `,` following the custom events example https://radar.com/documentation/api#body-parameters-2

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+- **`addressLabel`** (string, optional): The first line of the address, including the building number, street name, and unit number if applicable (e.g. 20 Jay St Apt 3).
+- **`city`** (string, required): The city name.
+- **`stateCode`** (string, required): The state abbreviation (e.g. NY).
```
jsani-radar, line 842: nit: `NY` and `US` are values so should be within code blocks

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+{
+  "meta": {
+    "code": 200
+  },
+  "address": {
+    "latitude": 40.7041,
+    "longitude": -73.9867,
+    "geometry": {
+      "type": "Point",
+      "coordinates": [-73.9867,40.7041]
```
jsani-radar, line 873: nit: space before second element

docs/api.mdx
```diff
@@ -811,6 +813,89 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
...
+    },
+    "country": "United States",
+    "countryCode": "US",
+    "county": "Kings County",
+    "city": "Brooklyn",
+    "number": "20",
+    "neighborhood": "DUMBO",
```
jsani-radar, line 880: nit: let's do this for the new office address

jsani-radar (commented):
few small comments, Bradley captured a lot of them - looks good overall

docs/api.mdx
bschoeneweis: Just one, but this is the current endpoint/what's in the spec

docs/api.mdx
bschoeneweis: That may be building/data specific.  Some definitely could have that, most don't though from what I've seen

jsani-radar: https://www.apartmentguide.com/blog/how-to-write-an-apartment-address/#pound-sign

bschoeneweis: > The pound sign (#) should not be used as a secondary unit designator if the correct designation, such as APT or STE, is known or is shown in the ZIP+4 file

source: https://pe.usps.com/text/pub28/28c2_003.htm

bschoeneweis: Seems like we'll be good without it, and if that designator is not found, I wonder if the source data will contain it anyways

jsani-radar: good to know, TIL

docs/api.mdx
palmakkar: yea good call - haven't seen a good pattern for 2 different examples in our docs. Would love to see any that I missed but I'm removing unit in the example here to keep it simple since the example request doesn't set expandUnits to true

docs/api.mdx
palmakkar: makes sense

docs/api.mdx
palmakkar: one address, but this is consistent though with the rest of our APIs (trips endpoints being /trips for example, even when you get/create/modify a single trip)

docs/api.mdx
palmakkar: yea works for me

docs/api.mdx
palmakkar: result is meant to explain how much we are able to verify the input address - I'd prefer to separate out this info and keep cleaned up address lean and consistent with how we represent it in other APIs. I could maybe see address being a field in result actually but don't have a strong opinion there/ Similar patterns found in Smarty and Google

docs/api.mdx
palmakkar: yea usps prefers unit designators over #

docs/api.mdx
palmakkar: nice catch - updated

docs/api.mdx
palmakkar: ah thanks for linking, didn't realize we have this pattern elsewhere. Updated to make it consistent

docs/api.mdx
palmakkar: good idea - probably worth updating the rest of our examples there

docs/api.mdx
```diff
@@ -811,6 +812,80 @@ curl "https://api.radar.io/v1/search/places?chains=starbucks&near=40.78382,-73.9
 }
 ```
 
+#### Validate an address
```
corypisano, line 815: wondering if we should make a `### Address` section? or we just want to keep under `### Search`. I think currently each section matches the url route (/geocode, /search)

palmakkar: yea it's a good question - went back and forth on this and discussed with Jay, I think it makes sense to add this to search as a starting point as ### Address would have only one endpoint, and can also lead to confusion if someone looking for address search jumps to the section and doesn't realize we do address search under ### Search. We can definitely revisit this once we grow our offering here

docs/api.mdx
```diff
@@ -580,6 +580,7 @@ Autocompletes partial addresses and place names, sorted by relevance.
 - **`layers`** (string, optional): Optional layer filters. A string, comma-separated, including one or more of `place`, `address`, `intersection`, `street`, `neighborhood`, `postalCode`, `locality`, `county`, `state`, `country`, `coarse`, and `fine`. Note that `coarse` includes all of `neighborhood`, `postalCode`, `locality`, `county`, `state`, and `country`, whereas `fine` includes all of `place`, `address`, `intersection`, and `street`. If not provided, results from all layers will be returned.
 - **`limit`** (number, optional): The max number of addresses to return. A number between 1 and 100. Defaults to 10.
 - **`country`** (string, optional): An optional countries filter. A string of comma separated countries, the unique 2-letter [country code](/regions/countries).
+- **`expandUnits`** (boolean, optional): If `true`, separate results will be returned for each unit in a building. Only available when `layers="address"` for the countries `US` and `CA`. Defaults to `false`. Currently in beta; ask your customer success manager for early access.
```
jsani-radar, line 583: S3 docs has an example of this: "While in beta, contact your customer success manager to enable this feature.

https://radar.com/documentation/integrations/aws-s3#configuration

jsani-radar (approved):
smallest nit, otherwise lgtm!

docs/api.mdx
palmakkar: Nice thanks for sharing, made the language more consistent. Keeping the "early access" language in here though, since this isn't really a feature that's ready to be enabled until April

PR #273, opened 2023-03-27
Update dwell events and times based on entry event

JoRoRadar:
## What?
Include the ramifications for turning on 'stop detection' as it relates to dwell events.

## Why?
With stop detection enabled, dwell time and dwell events are "sorta" incorrect as the dwell event timer starts when the geofence entry event is fired. The dwell time also starts counting when the geofence entry event fires.


docs/geofences.mdx
```diff
@@ -161,6 +161,10 @@ When _Geofence Stop Detection_ is on, Radar can understand the difference betwee
 
 Turn on _Geofence Stop Detection_ at a project level on the [Settings page](https://radar.com/dashboard/settings#geofences-settings).
 
+<Alert alertType="info">
+  Dwell Events and dwell time are generated in relation to the geofence entry event. With stop detection enabled, the geofence entry event will postpone the dwell event and reduce the total reported dwell time.
+</Alert>
```
jsani-radar, line 166: let's put this section under "Dwell events" below

also not sure if we should mention stop detection explicitly, or say that generating an entry event first is a requirement to start timing the dwell

doesn't feel right for us to explicitly call out that the metric is flawed vs outlining the potential limitations

jsani-radar (commented):
small comments, overall lgtm

docs/geofences.mdx
```diff
@@ -177,6 +173,10 @@ When _Buffer Exits_ is on, Radar will use the accuracy of the location reported
   Dwell events are currently in beta.
 </Alert>
 
+<Alert alertType="info">
+  Dwell events and dwell time require that the relevant geofence entry event be fired first. The dwell time will be calculated as the time since the geofence entry event up until the geofence exit event. The timer for firing the dwell event will also be in relation to the geofence entry event.
```
JoRoRadar, line 177: @jsani-radar - A little wordy but I like the explicitness.

jsani-radar (approved):

jsani-radar:
https://linear.app/radarlabs/issue/CORE-408/dwell-eventstiming-stop-detection

PR #278, opened 2023-03-30
Conversions

jsani-radar:


docs/api.mdx
```diff
@@ -2023,24 +2023,24 @@ To generate an entry or exit event, call the [track](#track) endpoint to trigger
...
 ###### Definition
 
 <span className="badge badge--info">POST</span> <code>https://api.radar.io/v1/events</code>
 
 ###### Body parameters
 
-- **`type`** (string, required): The type of the custom event. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
-- **`metadata`** (dictionary, optional): The metadata of the custom event. This data cannot contain any sensitive or personally identifiable information.
+- **`type`** (string, required): The type of the conversion event. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
+- **`metadata`** (dictionary, optional): The metadata of the conversion event. This data cannot contain any sensitive or personally identifiable information.
```
david-goodfellow, line 2043: we should include that `revenue` can be added to the metadata
also, are we going with the explicity `revenue` field? If so let's mention this and say it takes precedence over the `revenue` in metadata if both are given

docs/sdk/android.mdx
```diff
@@ -1497,12 +1496,11 @@ With the [custom events API](/api#send-a-custom-event), send a custom event, suc
   <TabItem value="kotlin">
 
   ```kotlin
-  Radar.sendEvent(
-      customType,
+  Radar.logConversion(
```
david-goodfellow, line 1499: should we have another one that includes revenue?

jsani-radar: good call, will add this across all SDKs

docs/api.mdx
jsani-radar: https://linear.app/radarlabs/issue/CORE-417/add-revenue-as-a-top-level-param-to-post-v1events

docs/api.mdx
```diff
@@ -2023,27 +2023,28 @@ To generate an entry or exit event, call the [track](#track) endpoint to trigger
...
 
-- **`type`** (string, required): The type of the custom event. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
-- **`metadata`** (dictionary, optional): The metadata of the custom event. This data cannot contain any sensitive or personally identifiable information.
+- **`type`** (string, required): The type of the conversion. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
+- **`metadata`** (dictionary, optional): The metadata of the conversion. This data cannot contain any sensitive or personally identifiable information.
 - **`userId`** (string, required): The stable unique ID for the user. Used to identify logged in users.
 - **`deviceId`** (string, optional): A device ID for the user. Used to identify logged out users.
-- **`createdAt`** (string, optional): The datetime when the event occurred.
+- **`revenue`** (number, optional): The revenue of the conversion. Used to associate an amount with a transactional conversion such as a purchase or coupon redemption.
+- **`createdAt`** (string, optional): The datetime when the conversion occurred.
```
david-goodfellow, line 2047: they can also technically give a `duration` as a number. This body param was originally specified within the OG custom events work

docs/api.mdx
```diff
@@ -2023,27 +2023,28 @@ To generate an entry or exit event, call the [track](#track) endpoint to trigger
...
 <span className="badge badge--info">POST</span> <code>https://api.radar.io/v1/events</code>
 
 ###### Body parameters
 
-- **`type`** (string, required): The type of the custom event. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
-- **`metadata`** (dictionary, optional): The metadata of the custom event. This data cannot contain any sensitive or personally identifiable information.
+- **`type`** (string, required): The type of the conversion. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
+- **`metadata`** (dictionary, optional): The metadata of the conversion. This data cannot contain any sensitive or personally identifiable information.
 - **`userId`** (string, required): The stable unique ID for the user. Used to identify logged in users.
 - **`deviceId`** (string, optional): A device ID for the user. Used to identify logged out users.
```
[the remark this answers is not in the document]
jsani-radar: so is the `userId` param optional as well? I don't want to overwhelm / confuse folks with all our ids - `id` is internal to us, and `installId` is an SDK impl detail

I assume sending a `userId` / `deviceId` that doesn't exist in our system would also register an opted-out conversion?

docs/api.mdx
```diff
@@ -2023,27 +2023,28 @@ To generate an entry or exit event, call the [track](#track) endpoint to trigger
...
 ###### Definition
 
 <span className="badge badge--info">POST</span> <code>https://api.radar.io/v1/events</code>
 
 ###### Body parameters
 
-- **`type`** (string, required): The type of the custom event. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
-- **`metadata`** (dictionary, optional): The metadata of the custom event. This data cannot contain any sensitive or personally identifiable information.
+- **`type`** (string, required): The type of the conversion. Should be named generically (e.g., `in_app_purchase`). Only alphanumeric characters and underscores are supported.
+- **`metadata`** (dictionary, optional): The metadata of the conversion. This data cannot contain any sensitive or personally identifiable information.
```
david-goodfellow, line 2043: I am assuming that we are tactfully not conflating `amount` with `revenue` here in the docs for simplicity? If so, still worth mentioning that `amount` is a reserved key to avoid bad instrumentation.

docs/api.mdx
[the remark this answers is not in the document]
david-goodfellow: `Missing user identifier. Must include one of id, userId, deviceId, or installId.`

If a `userId` or `deviceId` does not exist in our system, it will fail unless an `installId` is also provided:
```
const user = await UserFindService._findOneAsync({
      project: ctx.project,
      live: Boolean(ctx.live),
      id,
      userId,
      deviceId,
      installId,
      upsert: false,
    });

    if (!user && !installId) {
      throw new RadarHTTPError({
        message: 'Error finding user. For locationless conversion events, please provide an installId.',
        code: 404,
      });
    }
```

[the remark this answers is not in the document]
david-goodfellow: which is expected, we should not be polluting nonexistent user or device ids and we make the engagement stat calculated off distinct install id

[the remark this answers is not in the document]
jsani-radar: makes sense and understand the rationale for this decision - just seems a touch strange as I'm not sure customers will know what to pass for `installId`, and we don't doc it at all for `/track`

docs/api.mdx
jsani-radar: out of curiosity, why did we go with `amount` for the metadata key rather than `revenue` to be consistent? this is the ticket where we were discussing the new signature https://linear.app/radarlabs/issue/CORE-227/change-signature-of-custom-events-function-revenue-ios

was it because we had mentioned changing the signature to `amount` first?

david-goodfellow: yes we mentioned about before those signatures and it was already implemented on the server as such at that point in time

docs/api.mdx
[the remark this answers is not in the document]
david-goodfellow: yeah we automatically pass install id for the sdks. We can mention install id can technically be whatever if coming from a third party system as long as it is unique

david-goodfellow (approved):

docs/api.mdx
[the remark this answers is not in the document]
jsani-radar: I wonder if a more simple solve is to treat the user as opted-out if the `userId` / `deviceId` doesn't exist in Radar, though that'd be the case regardless when logging conversions for users before tracking hmm

feels like that'd be easier for customers rather than generating a random string and passing it thru for `installId`

PR #309, opened 2023-06-05
FENCE-487 Update Docs to reflect brand refresh

michaelchisari:
Change font colors, icon colors, alert info colors.


src/css/custom.css
```diff
@@ -43,22 +43,35 @@
...
+  --radar-gray-1: #F6FAFC;
+  --radar-gray-2: #EAF1F6;
+  --radar-gray-3: #DBE5EB;
+  --radar-gray-4: #C7D3DB;
+  --radar-gray-5: #ACBDC8;
+  --radar-gray-6: #5A6872;
+  --radar-gray-7: #32424D;
+  --radar-gray-8: #051723;
   --radar-gray-2-opaque: #e6edf25f; /* 30% opacity */
   --radar-gray-6-opaque: #173C561a; /* 10% opacity */
```
corypisano, line 55: wonder if these need a tweak

michaelchisari: Removed `gray-2-opaque` since the design says to use `gray-1` instead.

`gray-6-opaque` is only used for drop shadows, don't see that in figma, so that one might need to stay.

corypisano (approved):
Code looks good if we have a thumbs up from design 👍

PR #317, opened 2023-07-05
Add initialize for RN, bump play services version

jsani-radar:
## What?
https://linear.app/radarlabs/issue/FENCE-1007/react-native-initialization-docs-update

## Why?
https://app.slack.com/client/T1D3DQ9QA/C02MF6PAWBH/thread/C02MF6PAWBH-1687978893.979679/1688068275.433819

### DIFF
--- a/docs/sdk/react-native.mdx
+++ b/docs/sdk/react-native.mdx
@@ -58,9 +58,9 @@ Add the SDK to your project, preferably using CocoaPods. Finally, initialize the
 ```
 ### Android
 
-You must add Google Play Services Location to your project *(use version 18 or higher)*.
+You must add Google Play Services Location to your project *(use version 21.0.1 or higher)*.
 
-`implementation "com.google.android.gms:play-services-location:18.0.0"`
+`implementation "com.google.android.gms:play-services-location:21.0.1"`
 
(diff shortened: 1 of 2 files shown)


docs/sdk/react-native.mdx
```diff
@@ -99,6 +99,22 @@ First, import the module:
 import Radar from 'react-native-radar';
 ```
 
+### Initialize
+
+When your app starts, initialize the SDK with your publishable key.
+
+Use your `Test Publishable` key for testing and non-production environments. Use your `Live Publishable` key for production environments.
```
brettguenther, line 106: do we still need to initialize in the native code? Or is that only required if using background tracking w/o a blue bar/fg service (app wakes up from termination, we need to initialize)? If so, we should refactor/clean up the structure further.

jsani-radar: hmm I'm a bit confused, we do need to call `Radar.initialize` from RN to set the receiver such that the listeners actually receive locations and events - this applies regardless of tracking options to receive updates in the bg

https://github.com/radarlabs/react-native-radar/compare/3.5.7...3.7.4#diff-4ebacd111279fadadb2d49540af8037444202361fc44cdd79ebb4e23b97b1a49R80-R85

I'd prefer to get this in because the lack of initialization instructions in the RN docs prevents the listeners from working at all

brettguenther (approved):

PR #330, opened 2023-07-19
add roadClass values

gianiek:
added roadClass values

## What?
added roadClass values

## Why?
specifying what values will return 

## How?
confirmed values + added in docs

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1062,7 +1062,15 @@ Snaps points collected along a route to roads that were most likely traveled on.
 - **`path`** (string, required): List of up to 100 coordinates along a route to be snapped. A string in the format `latitude1,longitude1|latitude2,longitude2`.
 - **`mode`** (string, optional): The travel mode. A string with a value `foot`, `bike`, or `car`. Defaults to `car`.
 - **`roadAttributes`** (string, optional): Attributes of matched roads be included in the response. A string, comma-separated, including one or more of `speedLimit`, `names`, and `roadClass`.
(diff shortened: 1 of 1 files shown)


docs/api.mdx
```diff
@@ -1062,6 +1062,7 @@ Snaps points collected along a route to roads that were most likely traveled on.
 - **`path`** (string, required): List of up to 100 coordinates along a route to be snapped. A string in the format `latitude1,longitude1|latitude2,longitude2`.
 - **`mode`** (string, optional): The travel mode. A string with a value `foot`, `bike`, or `car`. Defaults to `car`.
 - **`roadAttributes`** (string, optional): Attributes of matched roads be included in the response. A string, comma-separated, including one or more of `speedLimit`, `names`, and `roadClass`.
+  - `roadClass` values: `motorway`, `trunk`, `primary`, `secondary`, `tertiary`, `unclassified`, `residential`, `service_other`
```
jsani-radar, line 1065: these are return values, not parameters correct?

if so we might want to do something similar to specifying the return values for address validation https://radar.com/documentation/api#validate-an-address

palmakkar: right, these shouldn't be in parameters - let's follow the address validation pattern as Jay suggested

gianiek: @palmakkar adjusted below - can update if any formatting nits!

palmakkar: thanks - just pushed a minor change unbolded the types, will merge it in now

palmakkar (approved):

PR #332, opened 2023-07-21
add directions API beta docs

palmakkar:

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1191,8 +1191,14 @@ curl "https://api.radar.io/v1/route/directions?locations=40.734250,-73.990927|40
     {
       "legs": [
         {
-          "startLocation": "40.734248,-73.990904",
-          "endLocation": "40.732934,-73.992442",
+          "startLocation": {
+            "latitude": 40.734248,
+            "longitude": -73.990904
+          },
+          "endLocation": {
+            "latitude": 40.732934,
+            "longitude": -73.992442
+          },
           "geometry": {
             "polyline": "mafulAnn`clChKxAhCF`CJdg@vGtB\\|BZpf@`HhC^yArEoZj`A"
           },
@@ -1206,8 +1212,14 @@ curl "https://api.radar.io/v1/route/directions?locations=40.734250,-73.990927|40
           }
         },
         {
-          "startLocation": "40.732934,-73.992442",
-          "endLocation": "40.73372,-73.994308",
+          "startLocation": {
+            "latitude": 40.732934,
+            "longitude": -73.992442
+          },
+          "endLocation": "startLocation": {
+            "latitude": 40.73372,
+            "longitude": -73.994308
+          },
           "geometry": {
             "polyline": "koculApncclCaTdq@aAzCqA~DmVpv@"
           },
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1150,10 +1150,10 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
 #### Directions
 
 <Alert alertType="info">
-  The directions API is currently in beta.
+  The directions API is currently in beta. Contact your customer success manager for early access.
 </Alert>
 
-Calculates the most efficient route between two or more locations.
+Calculates the most efficient route between two or more locations to visit in order.
 
 ###### Definitions
 
@@ -1161,9 +1161,9 @@ Calculates the most efficient route between two or more locations.
 
 ###### Query parameters
 
-- **`locations`** (string, required): List of up to 25 coordinates to visit sequentially. A pipe-delimited string in the format `latitude0,longitude0|latitude1,longitude1|...`.
+- **`locations`** (string, required): List of up to 25 coordinates to visit in order. A pipe-delimited string in the format `latitude0,longitude0|latitude1,longitude1|...`.
 - **`mode`** (string, optional): The travel mode. A string with a value `foot`, `bike`, or `car`. Defaults to `car`.
-- **`units`** (string, optional): The distance units. A string, `metric` or `imperial`. Defaults to `imperial` if not provided.
+- **`units`** (string, optional): The distance units. A string, `metric` or `imperial`. Defaults to `imperial`.
 
 ###### Authentication level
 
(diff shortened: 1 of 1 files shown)


docs/api.mdx
```diff
@@ -1147,6 +1147,68 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
 }
 ```
 
+#### Directions
+
+<Alert alertType="info">
+  The directions API is currently in beta.
+</Alert>
+
+Calculates the most efficient route between two or more locations.
```
jsani-radar, line 1156: this makes it seem like the endpoint does route optimization as well - should be more explicit about providing directions if that's the core purpose of the API

docs/api.mdx
```diff
@@ -1147,6 +1147,68 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
...
+
+###### Definitions
+
+<span className="badge badge--success">GET</span> <code>https://api.radar.io/v1/route/directions</code>
+
+###### Query parameters
+
+- **`locations`** (string, required): List of up to 25 coordinates to visit sequentially. A pipe-delimited string in the format `latitude0,longitude0|latitude1,longitude1|...`.
+- **`mode`** (string, optional): The travel mode. A string with a value `foot`, `bike`, or `car`. Defaults to `car`.
+- **`units`** (string, optional): The distance units. A string, `metric` or `imperial`. Defaults to `imperial` if not provided.
```
jsani-radar, line 1166: nit: remove `if not provided` as it's not present in the above line for the `mode` default

docs/api.mdx
```diff
@@ -1147,6 +1147,68 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
...
+      "duration": {
+        "value": 2.1667,
+        "text": "2 mins"
+      }
+    }
+  }
+}
+```
+
+While in beta, contact your customer success manager for early access to the directions API.
```
jsani-radar, line 1210: why not put this in the info alert up top?

docs/api.mdx
```diff
@@ -1147,6 +1147,68 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
...
+Publishable
+
+###### Default rate limit
+
+10 requests per second
+
+###### Sample request
+
+```bash
+curl "https://api.radar.io/v1/route/directions?locations=40.73421,-73.98635|40.73616,-73.98026&mode=car&units=imperial" \
```
jsani-radar, line 1179: wondering if we should provide an example with 3 coordinates instead of just 2 - otherwise it's basically just a call to the distance API

docs/api.mdx
```diff
@@ -1147,6 +1147,68 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
...
+
+###### Sample response
+
+```json
+{
+  "meta": {
+    "code": 200
+  },
+  "routes": {
+    "legs": {
```
jsani-radar, line 1191: should `legs` be an array? if there are multiple legs you can't have redundant `startLocation` and `endLocation` keys in the same object

jsani-radar (commented):
small nits, overall looking good

docs/api.mdx
palmakkar: thanks for flagging - I can see the potential confusion with optimization API. what this is getting at is this is the "best" route in terms of travel time as others like Mapbox point out in their docs but going to wordsmith this a bit

It should evolve into a  full blown turn by turn directions API but note we're not including turn by instructions yet -  those need a bit more scoping but we need to rush out the route polyline functionality for multiple points for Pilot as the V1

docs/api.mdx
palmakkar: good idea - moved it up

docs/api.mdx
bschoeneweis: following up here, I might even remove "the most efficient" and just say "Calculates a route between two or more locations."

vivekjain1397 (approved):
lgtm

docs/api.mdx
```diff
@@ -1147,6 +1147,85 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
...
+```json
+{
+  "meta": {
+    "code": 200
+  },
+  "routes": [
+    {
+      "legs": [
+        {
+          "startLocation": "40.734248,-73.990904",
```
bschoeneweis, line 1194: this is sort of weird, we should return either a coordinates array `[<longitude>, <latitude>]` or a geometry point:

```
{
        "type": "Point",
        "coordinates": [<longitude>, <latitude>]
}
```

docs/api.mdx
palmakkar: I updated this to "between two or more locations" which should make it explicit -- think it's worth communicating we're choosing an optimal route before drawing it out (common customer question for distance API is how is the route chosen). We can also revisit before getting it out of beta

docs/api.mdx
```diff
@@ -1147,6 +1147,85 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
...
+          "endLocation": "40.732934,-73.992442",
```
bschoeneweis, line 1195: same for end

docs/api.mdx
```diff
@@ -1147,6 +1147,85 @@ curl "https://api.radar.io/v1/route/match?path=40.73421259605101,-73.98635368448
...
+            "text": "0.2 mi"
+          },
+          "duration": {
+            "value": 0.6890000000000001,
+            "text": "1 min"
+          }
+        },
+        {
+          "startLocation": "40.732934,-73.992442",
+          "endLocation": "40.73372,-73.994308",
```
bschoeneweis, line 1210: same as above

docs/api.mdx
palmakkar: yea  that's what we use in geocoding but that feels like a lot of nesting and we have another geometry field below for polyline. Within routing endpoints, I see more of the 
{
      "latitude": 40.70390,
      "longitude": -73.98670
    }
format across matrix and route matching

going to go with that for the beta

PR #345, opened 2023-08-01
Android maps docs with Maplibre

jaspk06:
## What?

Adding missing docs for native Android maplibre

## Screenshots (optional)

Screen_Recording_20230801_150713_Getting Started_1

### DIFF
--- a/docs/maps/maps.mdx
+++ b/docs/maps/maps.mdx
@@ -397,13 +397,19 @@ class ViewController: UIViewController, MGLMapViewDelegate {
 }
 ```
 
+You can find the assets used for the pin and Radar logo [here](/img/radar-map-assets.zip).
+
+<Alert alertType="info">
+  Logo placement is required as defined in the <a href="https://radar.com/terms">Terms of Use</a>.
+</Alert>
+
 ### Android
 
 Radar maps can be integrated into an Android application with [MapLibre Native](https://github.com/maplibre/maplibre-native).
 
 To get started, add the MapLibre Native dependency into your module Gradle file (usually located at `<project>/<app-module>/build.gradle`). Replace `<version>` with the latest MapLibre Native version (e.g.: `org.maplibre.gl:android-sdk:10.2.0`).
 
-```kotlin title="build.gradle"
+```kotlin
 dependencies {
     ...
     implementation 'org.maplibre.gl:android-sdk:<version>'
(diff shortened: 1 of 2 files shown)


docs/maps/maps.mdx
```diff
@@ -397,6 +397,176 @@ class ViewController: UIViewController, MGLMapViewDelegate {
 }
 ```
 
+### Android
+
+Radar maps can be integrated into an Android application with [MapLibre Native](https://github.com/maplibre/maplibre-native).
+
+To get started, add the MapLibre Native dependency into your module Gradle file (usually located at `<project>/<app-module>/build.gradle`). Replace `<version>` with the latest MapLibre Native version (e.g.: `org.maplibre.gl:android-sdk:10.2.0`).
+
+```kotlin title="build.gradle"
```
palmakkar, line 406: I like these file names for clarity, but given we don't use them anywhere else in docs yet let's just leave it out for consistency

docs/maps/maps.mdx
```diff
@@ -397,6 +397,176 @@ class ViewController: UIViewController, MGLMapViewDelegate {
...
+            .fromBitmap(bitmapMarker)
+
+        // create point to place marker
+        val markerOptions = MarkerOptions()
+            .position(coordinate)
+            .icon(icon)
+        mapboxMap.addMarker(markerOptions)
+    }
+
+    override fun onStart() {
```
palmakkar, line 533: should we leave out any boilerplate code and keep the snippets relevant to this step? think that would be more consistent with the way we've documented the other platforms

docs/maps/maps.mdx
jaspk06: the issue with android is that there's multiple config files with the same name like `build.gradle` where it appears in 3 different directories so I feel like it still needs to be included somehow. I agree with the other code blocks tho, the file names are unnecessary

docs/maps/maps.mdx
jaspk06: the boilerplate is kinda necessary in this case cuz on each lifecycle call like `onStart()` `mapView.onStart()` also needs to be called.

jaspk06: which isn't already included in the default boilerplate

Gdakgdak22 (approved):

lmeier:
FYI we're using this map logo on React Native, per @berylw's instruction: 
image

https://github.com/radarlabs/react-native-radar/blob/master/js/ui/map-logo.png

kochis:
I think we should be using the same maps logo everywhere for consistency. The darker one should be the correct one.

palmakkar (approved):

Gdakgdak22 (approved):

PR #360, opened 2023-08-09
remove coordinates from airship integration

brettguenther:
## What?

Associated with FENCE-1163 - airship tag removal for coordinates

## Why?

Can't be used in Airship integration, so removing from payloads sent (would need to sent as new user attributes feature)


jsani-radar:
https://linear.app/radarlabs/issue/FENCE-1163/stop-sending-latitude-and-longitude-as-airship-channel-tags

jsani-radar (approved):

PR #361, opened 2023-08-10
Initial Commit

JoRoRadar:
@jsani-radar - I see we use Express to handle our rate limits so it should be standard across all our API's but I did notice that:
1) /track always returns these headers while other API's don't
2) Whe diagnosing issues for GPI on beacons we were also receiving a `type` field [sec, min, hr, day]. I think our server appends extra rate limit info but unsure how much we want to potentially include here.


docs/geofencing/testing.mdx
```diff
@@ -80,3 +80,32 @@ The Radar [iOS](/sdk/ios) and [Android](/sdk/android) SDKs automatically generat
 ### Rate limit errors are occurring when tracking location
 
 Radar has device specific rate limits per second, hour and day for the [track API endpoint](/api#track), which sends location updates to Radar. To mitigate rate limits, adjust [tracking options](/sdk/tracking) to collect locations less frequently. If you hit the per hour or day limits while testing, reinstalling the application with the Radar SDK will reset the rate limits.
+
+When a rate limit is triggered, Radar will return three response headers for the request inline with the IETF Rate Limiting Standardization Draft:
+
+###### Response codes
+
+- **`x-ratelimit-limit`**: Maximum amount of request allowed in the current window.
+- **`x-ratelimit-remaining`**: Remaining requests available in the current window before being rate limited.
```
brettguenther, line 89: can we make sure to match casing to the sample response?

PR #369, opened 2023-08-22
add units parameter to route match query

palmakkar:


corypisano (approved):
Looks good, just while we're here noticed in roadAttributes we're missing a "to" in 
`Attributes of matched roads be included in the response`

PR #389, opened 2023-10-09
add truck travel mode

palmakkar:
## What?

add truck travel mode to docs

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -677,7 +677,7 @@ Searches for users near a location, sorted by distance.
 
 - **`near`** (string, required): A location for the search. A string in the format `latitude,longitude`.
 - **`radius`** (number, optional): The radius to search, in meters. A number between 100 and 10000. Defaults to 1000. If `mode` is specified, the `radius` is the travel duration in minutes.
-- **`mode`** (string, optional): The [travel mode](/api#route-distance). A string, one of `car`, `truck`, `foot`, and `bike`.
+- **`mode`** (string, optional): The [travel mode](/api#distance). A string, one of `car`, `truck`, `foot`, or `bike`.
 - **`metadata[key]`** (string, optional): Optional metadata filters. Values may be of type string. Type will be automatically inferred. For example, to match on `working == true`, use `&metadata[working]=true`.
 - **`limit`** (number, optional): The max number of users to return. A number between 1 and 100. Defaults to 10.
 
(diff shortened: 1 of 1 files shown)


docs/api.mdx
```diff
@@ -677,7 +677,7 @@ Searches for users near a location, sorted by distance.
 
 - **`near`** (string, required): A location for the search. A string in the format `latitude,longitude`.
 - **`radius`** (number, optional): The radius to search, in meters. A number between 100 and 10000. Defaults to 1000. If `mode` is specified, the `radius` is the travel duration in minutes.
-- **`mode`** (string, optional): The [travel mode](/api#route-distance). A string with a value of `foot`, `bike`, or `car`.
+- **`mode`** (string, optional): The [travel mode](/api#route-distance). A string, one of `car`, `truck`, `foot`, and `bike`.
```
jsani-radar, line 680: the `#route-distance` href doesn't take me anywhere - think it might be just `#distance` now?
Screenshot 2023-10-09 at 1 27 03 PM

docs/api.mdx
```diff
@@ -941,7 +941,7 @@ Calculates the travel distance and duration between an origin and a destination.
 
 - **`origin`** (string, required): The origin. A string in the format `latitude,longitude`.
 - **`destination`** (string, required): The destination. A string in the format `latitude,longitude`.
-- **`modes`** (string, required): The [travel modes](/api#route-distance). A string, comma-separated, including one or more of `foot`, `bike`, and `car`.
+- **`modes`** (string, required): The travel modes. A string, comma-separated, including one or more of `car`, `truck`, `foot`, and `bike`.
```
jsani-radar, line 944: know we have "and" everywhere in the API ref, but wondering if it should be "or" since it's "one of"

thoughts?

docs/api.mdx
```diff
@@ -1198,7 +1198,7 @@ Calculates the most efficient route between two or more locations to visit in or
 ###### Query parameters
 
 - **`locations`** (string, required): A list of up to 25 coordinates to visit in order. A pipe-delimited string in the format `latitude0,longitude0|latitude1,longitude1|...`.
-- **`mode`** (string, optional): The travel mode. A string with a value `foot`, `bike`, or `car`. Defaults to `car`.
+- **`mode`** (string, optional): The travel mode. A string, one of `car`, `truck`, `foot`, and `bike`. Defaults to `car`.
```
jsani-radar, line 1201: doesn't seem like we have "Defaults to `car`" everywhere consistently - let's make sure that appears everywhere it's applicable, think that's wherever it's referred to as a query param

also makes sense to have "travel mode" be consistently a link everywhere

jsani-radar (commented):
small comments - mostly around consistency and some light cleanup

otherwise lgtm and will be 👍🏽 on next pass

docs/api.mdx
palmakkar: good call this looks out of date, fixing it

docs/api.mdx
palmakkar: makes sense 👍

docs/api.mdx
palmakkar: I think the places where we don't have default noted are when mode is required, so there's no applicable default

the link to distance api doesn't feel like it adds helpful context (it's the same blurb) - search users might be an exception since it's using distance under the hood but if anything for consistency, I'd be a fan of removing it everywhere

jsani-radar: makes sense if it's repeated everywhere anyways - feel free to remove

PR #418, opened 2023-12-08
Maps and Static Maps updates

kochis:
### Fit map to markers example
image

### Add custom icon example
image

### Change map language example
image

### Paths section
image

### Add custom icon example
image

### Route example
image

Updates to static maps:
* Add `path` drawing section


gregsadetsky:
found a small issue with the `fitToMarkers` where clicking a few times will show a map with only the most recent markers on the map. clicking again will show the map with all of the markers, then clicking again will show a map with not all markers

https://www.loom.com/share/ff1d1969d06d459dbcf728fadc85f1fc

gregsadetsky (approved):

PR #428, opened 2024-01-04
Fix links to CLVisit

jsani-radar:
fix links pointing to stale / incorrect URLs and anchor tags


jsani-radar:
https://linear.app/radarlabs/issue/RAD-5864/i-noticed-that-the-link-under-the-verified-[fraud

PR #431, opened 2024-01-11
Add docs for GET /trips/:tripId/route

bschoeneweis:
## What?
API docs for GET /trips/:tripId/route

## Screenshots (optional)
Screenshot 2024-01-11 at 12 25 29 PM

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1879,7 +1879,7 @@ Secret
   Contact your customer success manager to enable this feature.
 </Alert>
 
-Gets a [route match](#route-match) for a completed, canceled, or expired trip. The trip can be referenced by Radar `_id` or `externalId`. The trip must have at least two locations for a route match to be calculated.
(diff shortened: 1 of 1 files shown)


docs/api.mdx
```diff
@@ -1873,6 +1873,49 @@ Secret
 }
 ```
 
+#### Get the route for a trip
+
+<Alert alertType="info">
+  Contact your customer success manager to enable this feature.
+</Alert>
+
+Gets a [route match](#route-match) for a completed, canceled, or expired trip. The trip can be referenced by Radar `_id` or `externalId`. The trip must have at least two locations for a route match to be calculated.
```
kochis, line 1882: Maybe `Gets a route match response` or `Gets a matched route`

kochis (approved):

docs/api.mdx
bschoeneweis: i'll go with `Gets a matched route` since the response shapes look a little different

PR #438, opened 2024-02-07
fix curl example in api.mdx

gregsadetsky:
spaces after commas otherwise lead curl to report "curl: (3) URL rejected: Malformed input to a URL function"

actually, the API doesn't accept that string with spaces either -- it returns "roadAttributes: Invalid string array. Valid values: names, speedLimit, roadClass" if spaces are passed

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1121,7 +1121,7 @@ Publishable
 ###### Sample request
 
 ```bash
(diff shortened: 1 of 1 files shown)


kochis (approved):
Good catch, thanks!

tjulien (approved):

PR #439, opened 2024-02-14
waypoint doc

KennyHuRadar:
**NOTE** The links are not yet updated(still points to toolkit download links) and should not be linked to the side bar yet. But this page is needed for app store review.

New page can be viewed @ `/waypoint` path


lmeier:
2 things:
- until we have this on the app store, let's remove the links to the app store (which now confusingly go to the Toolkit pages)
- let's put a "Help" section at the bottom indicating that if they have any issues, to email support@radar.com

lmeier:
The navigation sidebar doesn't show on the page for me as well presently. Can we keep the sidebar even if we're not linking to this page from it?

lmeier (approved):

PR #452, opened 2024-03-13
Trip olo metadata for reliable order firing

corypisano:
## What?

Make including the oloOrderGuid and oloHandoffMode as trip metadata explicit in Olo integration docs for best order firing reliability

## Why?

Now a default recommendation to customers and should be explicit in our docs

## Screenshots (optional)

image


ahwreck (approved):
looks good 👍

tjulien (approved):

PR #463, opened 2024-04-01
Update flutter and android sdk to use fetched value from GH

KennyHuRadar:


tjulien:
@KennyHuRadar looks like we might have added this in another PR since this was opened?

PR #488, opened 2024-05-31
Update Android example in maps docs

jaspk06:
https://documentation-git-update-maps-android-radarlabs.vercel.app/documentation/maps/maps#android

## What?
Updates Android map example

## Why?
Current example uses deprecated patterns like `addMarker`

## How?
Changed to use their recommended plugin pattern
see threads:
https://github.com/maplibre/maplibre-native/issues/2140
https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md

## Screenshots (optional)
example in app:
image

### DIFF
--- a/docs/maps/maps.mdx
+++ b/docs/maps/maps.mdx
@@ -448,7 +448,7 @@ To create a map on Android, add [MapLibre Native](https://github.com/maplibre/ma
 ```gradle
 dependencies {
     implementation 'org.maplibre.gl:android-sdk:11.0.0'
-    implementation 'org.maplibre.gl:android-plugin-annotation-v9:3.0.0'
+    implementation 'org.maplibre.gl:android-plugin-annotation-v9:3.0.0' // optional import for adding a marker
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/maps/maps.mdx
```diff
@@ -525,40 +527,51 @@ class MainActivity : AppCompatActivity() {
...
-            map.cameraPosition = CameraPosition.Builder().target(LatLng(40.7342, -73.9911)).zoom(11.0).build()
+            map.setStyle(styleURL) {style ->
+
+                // add a marker to the map on style load
+                val infoIconDrawable = ResourcesCompat.getDrawable(
+                    this.resources,
+                    // use imported marker resource
+                    R.drawable.default_marker,
+                    null
+                )!!
```
lmeier, line 547: what's this `!!`?

Let's add a comment indicating that this part specifically requires the optional marker dep

jaspk06: it's a non null assertion since `ResourcesCompat.getDrawable` can be `Nullable` [(source)](https://developer.android.com/reference/androidx/annotation/Nullable#getDrawable(android.content.res.Resources,int,android.content.res.Resources.Theme))

however, this is just for safety in case the resource doesn't exist. In our example, we specify the asset we're using. Without `!!` we'd explicitly have to write some error handling which is a bit verbose for an example

lmeier (approved):

PR #501, opened 2024-06-12
Update fraud.mdx fix trackVerified related signatures

ShiCheng-Lu:

### DIFF
--- a/docs/fraud.mdx
+++ b/docs/fraud.mdx
@@ -314,7 +314,8 @@ Radar.trackVerified().then((result) => {
   <TabItem value="react-native">
 
 ```javascript
-Radar.trackVerified().then(({ status, token }) => {
+Radar.trackVerified().then((result) => {
+  const { status, token } = result;
   if (token?.passed) {
     // allow access to feature, send token to server
   } else {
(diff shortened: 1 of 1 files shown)


docs/fraud.mdx
```diff
@@ -314,7 +314,7 @@ Radar.trackVerified().then((result) => {
   <TabItem value="react-native">
 
 ```javascript
-Radar.trackVerified().then((result) => {
+Radar.trackVerified().then(({ status, result }) => {
```
lmeier, line 317: Hmmm, should we make param `token` so that this would be `status, token` like it is in swift, kotlin, etc?

nickpatrick:
Thx for this. Two things:

(1) Looks like there's an extra `(` in the callback

(2) Moot point because I would prefer that we keep this consistent with the `trackOnce()` docs (https://radar.com/documentation/sdk/react-native#foreground-tracking). So

```
Radar.trackVerified().then((result) => {
  const { token } = result;
  if (token?.passed) {
    // allow access to feature, send token to server
  } else {
    // deny access to feature, show error message
  }
}).catch((err) => {
  // deny access to feature, show error message
});
```

lmeier (approved):

PR #520, opened 2024-07-17
iOS trips fg service description update

gianiek:
adds dynamic island description to location indicator note in trips docs

## What?

adds dynamic island description to location indicator note in trips docs

## Why?
includes newest fg service location indicator experience 

## How?
clarified phrasing 

## Anything Else? (optional)
from slack discussion here: https://radarlabs.slack.com/archives/C06V789S55F/p1721167958241539


nickpatrick (approved):

nickpatrick:
Awesome, thx. Nit, but we might want to capitalize Dynamic Island, seems like Apple does that everywhere

gianiek:
> Awesome, thx. Nit, but we might want to capitalize Dynamic Island, seems like Apple does that everywhere

ah thanks, updated!

PR #528, opened 2024-08-01
Update dmas.mdx

ahwreck:
fixing incomplete dma names per schemas/regions/dmas.json

### DIFF
--- a/docs/regions/dmas.mdx
+++ b/docs/regions/dmas.mdx
@@ -19,31 +19,31 @@ Code | Name
 `508` | `Pittsburgh`
 `509` | `Fort Wayne`
 `510` | `Cleveland-Akron`
-`511` | `Washington`
+`511` | `Washington, D.C.`
 `512` | `Baltimore`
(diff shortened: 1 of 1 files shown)


docs/regions/dmas.mdx
```diff
@@ -19,31 +19,31 @@ Code | Name
...
 `510` | `Cleveland-Akron`
-`511` | `Washington`
+`511` | `Washington, D.C.`
 `512` | `Baltimore`
 `513` | `Flint-Saginaw-Bay City`
 `514` | `Buffalo`
 `515` | `Cincinnati`
 `516` | `Erie`
-`517` | `Charlotte`
+`517` | `Charlotte, SC`
```
tjulien, line 28: we should probably add the state to every DMA for consistency but that can come later

tjulien, line 28: we should probably add the state to every DMA for consistency but that can be a follow up

tjulien (approved):

PR #529, opened 2024-08-01
adding borderCrossings as a param to avoid for matrix endpoint

vivekjain1397:

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -1069,7 +1069,7 @@ Calculates the travel distances and durations between multiple origins and desti
 - **`destinations`** (string, required): A list of destinations. A pipe-delimited string in the format `latitude0,longitude0|latitude1,longitude1|...`.
 - **`mode`** (string, required): The travel mode. A string, one of `car`, `truck`, `foot`, or `bike`.
 - **`units`** (string, optional): The distance units. A string, `metric` or `imperial`. Defaults to `imperial` if not provided.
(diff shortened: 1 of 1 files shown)


corypisano:
avoid borderCrossings only added to distance and directions API initially

docs/api.mdx
```diff
@@ -1069,7 +1069,7 @@ Calculates the travel distances and durations between multiple origins and desti
 - **`destinations`** (string, required): A list of destinations. A pipe-delimited string in the format `latitude0,longitude0|latitude1,longitude1|...`.
 - **`mode`** (string, required): The travel mode. A string, one of `car`, `truck`, `foot`, or `bike`.
 - **`units`** (string, optional): The distance units. A string, `metric` or `imperial`. Defaults to `imperial` if not provided.
-- **`avoid`** (string, optional): The features that calculated routes should avoid. A string, comma-separated, including one or more of `tolls`, `highways` and `ferries`. No features are avoided if not provided.
+- **`avoid`** (string, optional): The features that calculated routes should avoid. A string, comma-separated, including one or more of `tolls`, `highways`, `ferries`, `borderCrossings`. No features are avoided if not provided.
```
tjulien, line 1072: @corypisano thinking maybe `borders` would be simpler and more consistent with the other params 🤔 .  If you agree I can make the docs and code change as well (but leave the old one for back compat) cc @vivekjain1397

PR #580, opened 2024-12-09
New updated Unity SDK documentation

PythonHackerr:
Unity SDK documentation

### DIFF
--- a/docs/sdk/unity.mdx
+++ b/docs/sdk/unity.mdx
@@ -8,279 +8,67 @@ sidebar_position: 1
 
 This is the documentation for the Unity wrapper/plugin, which provides a thin proxy layer for integrating existing native SDKs offering location tracking and geofencing services. Before integrating, review the native SDK documentation for iOS and Android to understand platform-specific requirements and features.
 
-The Unity wrapper is developed as a plugin for native SDKs written in Kotlin for Android and Objective-C for iOS.
-
 See the source on GitHub [here](https://github.com/radarlabs/unity-radar).
 
-## Supported Unity versions
-
-The SDK has been tested on Unity **2022.3.41f**, and **6000.0.24f** versions. However, it should work with Unity versions **2021** and above.
-
-## Project structure
-
-<div style={{ display: "flex", alignItems: "flex-start", gap: "20px" }}>
-  <div style={{ flex: "1.5" }}>
-    <p>The project directory is organized as follows:</p>
-  </div>
-  <div style={{ flex: "1" }}>
-    <img src="https://images.ctfassets.net/f2vbu16fzuly/1g09wPuKjYu2YH622UTD2y/d080966101b93adbea2f8e4745a5f97f/Image4.png" alt="Demo Scene Status Lights" style={{ width: "75%", height: "auto" }} />
-  </div>
-</div>
-
-
-### Scripts overview
-
-This section provides an overview of the main scripts in the Radar SDK for Unity, detailing each script's role and its relationships with other components. This will help you quickly understand how to integrate and utilize the Radar SDK in your Unity project.
-
-#### 1. `RadarExample.cs`
-
-> **Description**:  
-> Main script used in demo scene. Demonstrates how to initialize and configure the Radar SDK in Unity. This script provides a basic setup and can be used as a starting template for integrating Radar SDK functionality.
-
-| **Functionality**                                                                              | **Related Scripts**                                                                                  |
-|------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
-| - Initializes the Radar SDK using `RadarSettingsData`.                                         | - Works with `RadarServiceWrapper` to perform core SDK operations.                                   |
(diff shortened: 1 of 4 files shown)


docs/sdk/unity.mdx
```diff
@@ -0,0 +1,408 @@
...
+using RadarSDK;
+```
+
+
+
+
+
+### Calls to APIs
+
+The Unity SDK exposes following APIs calls:
```
KennyHuRadar, line 222: I don't believe having a comma separated list is the best way to index our api calls, maybe a codebox with pseudo code in a class format looks better? Like the C# version of
```
RadarSDK {
initialize
...
}
```

And also since we already describe the calls below, do we even need to list them out here?

docs/sdk/unity.mdx
```diff
@@ -0,0 +1,408 @@
...
+```cs
+private async Task StopTrackingAsync()
+{
+   await Radar.StopTracking();
+}
+```
+
+#### GetVerifiedLocationToken
+
+Once the native SDK retrieves the location, it sends the data back to Unity. On Android, this is done via RadarJavaProxy, and on iOS, it’s done via a delegate or callback defined in RadarUnityBridge.m
```
KennyHuRadar, line 308: we should either have a descriptions for all api calls or for none of them for consistency.

docs/sdk/unity.mdx
```diff
@@ -0,0 +1,408 @@
...
+See the [Fraud Error Handling Documentation](https://radar.com/documentation/fraud#error-handling) for detailed error messages and resolutions.
+
+### Example Integration
+
+Here’s an example of how to log errors using the SDK:
+
+```csharp
+try
+{
+   Radar.SomeFunctionality();
```
KennyHuRadar, line 396: let's use a real call here

package.json
```diff
@@ -49,6 +49,7 @@
     "@types/react": "^17.0.11",
     "@types/react-helmet": "^6.1.1",
     "@types/react-router-dom": "^5.1.7",
+    "cross-env": "^7.0.3",
```
KennyHuRadar, line 52: why are we using cross env? is this just a local dev dependency you need?

Description
KennyHuRadar, line 1: why do we need this? is this also related to the cross-env?

docs/sdk/unity.mdx
```diff
@@ -0,0 +1,408 @@
...
+
+# Unity wrapper SDK
+
+This is the documentation for the Unity wrapper/plugin, which provides a thin proxy layer for integrating existing native SDKs offering location tracking and geofencing services. Before integrating, review the native SDK documentation for iOS and Android to understand platform-specific requirements and features.
+
+See the source on GitHub [here](https://github.com/radarlabs/unity-radar).
+
+Scripts documentation and Demo scene overview can be found in project READMEs.
+
+## Install
```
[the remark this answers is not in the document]
PythonHackerr: We can and we should, I will create it and publish to the repo

docs/sdk/unity.mdx
```diff
@@ -0,0 +1,408 @@
...
+
+To integrate the Unity SDK, download or clone the package and include it in your Unity project:
+
+1. Copy the `Plugins` folder (containing both Android and iOS native implementations) into the `Assets` directory of your Unity project. Ensure it contains all the required files.
+2. Import the ExternalDependencyManager for Unity (EDM4U). It is included in the project but you can get it from `Package Manager` or from [repo](https://github.com/googlesamples/unity-jar-resolver): It is required if you want to add and use iOS/Android dependencies directly in your project like: `Android specific libraries (e.g AARs)` or `iOS CocoaPods`
```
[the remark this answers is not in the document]
PythonHackerr: We should!

Description
PythonHackerr: What exactly?

package.json
PythonHackerr: I think we do not need this. Perhaps I added this when I tried to fix some issues. I will delete it and test everything.

docs/sdk/unity.mdx
PythonHackerr: I will add some short descriptions that make sense to all functions

docs/sdk/unity.mdx
PythonHackerr: Agree, no need to list them out

package.json
PythonHackerr: Without it, I cannot preview. I get this error:

```
$ npm run start

> docs@0.0.0 start
> NODE_OPTIONS=--openssl-legacy-provider docusaurus start

'NODE_OPTIONS' is not recognized as an internal or external command,
operable program or batch file.
```

KennyHuRadar: I see, are you working on a windows PC?

Description
KennyHuRadar: I'm curious on why do we need to add this new file into the repository, this wasn't previously here.

KennyHuRadar:
The preview is not building, I suspect that there may be an issue with the new files you created. I am guessing that you need these files to build the file for you but they are breaking the actual deployment build process. Can you remove those file diffs from this PR, you can obviously keep those on your local copy of the repo but I think we don't want to merge them in.

PR #588, opened 2025-01-16
Update SSO Docs

JoRoRadar:
"Assertion Consumer Service URL" and "Application Callback URL" are the same. The way the sentence is written, and the items that Vail asked for, were both these URLs when in reality there is only one. Removed the "Application Callback URL" as the ACS URL is the more standard SAML terminology.

In its place, I added "Entity ID" as this is something that they will also need.


JoRoRadar:
@lmeier - I got a couple of these docs prs open.

PR #598, opened 2025-02-14
Update olo.mdx to reflect new manual firing behavior

gregsadetsky:

### DIFF
--- a/docs/integrations/olo.mdx
+++ b/docs/integrations/olo.mdx
@@ -39,7 +39,7 @@ Assuming locations permissions have been granted and location tracking has been 
 
 Radar can fire orders in Olo based on customer location rather than a preset time to ensure food preparation time aligns with customer arrival.
 
-Orders in Olo will need to have `ismanualfire` set to `true` or set a `timewanted` for advance orders.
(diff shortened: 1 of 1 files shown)


docs/integrations/olo.mdx
```diff
@@ -39,7 +39,7 @@ Assuming locations permissions have been granted and location tracking has been
 
 Radar can fire orders in Olo based on customer location rather than a preset time to ensure food preparation time aligns with customer arrival.
 
-Orders in Olo will need to have `ismanualfire` set to `true` or set a `timewanted` for advance orders.
+Orders in Olo will need to set a `timewanted` for advance orders, or alternatively, set the `externalRef` to any value that starts with `radar_` (i.e. `radar_h87y12rgb`).
```
lmeier, line 42: @JoRoRadar — in the slack thread you mentioned that this `externalRef` functions to turn `isManualFire` true, and this behavior would happen _**instead**_ of changing that metadata.

Presently `externalRef` works as an override, and one can still use that metadata. I imagine some people are using the metadata still, right? If so, maybe we'd want to keep a reference to that approach in the docs here?

lmeier (approved):

PR #599, opened 2025-02-14
consolidate campaign and notification docs

KennyHuRadar:


KennyHuRadar:
There is also one tutorial that demonstrates an implementation that is rendered obsolete by campaigns, should we delete that as well? Furthermore, should we create a campaigns tutorial?

lmeier (approved):

PR #606, opened 2025-03-27
Add "Display a globe" section to docs

siddie-pennewill:
## What?

"Display a globe" example in Maps docs to add guidance for customers. 

## Why?

Maersk (enterprise deal) wanted to see that Radar Maps can be rendered as a globe in order to move forward. This section will guide them through implementation and answer their main request.

## How?

Used https://maplibre.org/maplibre-gl-js/docs/examples/globe-vector-tiles/ as guidance.

## Screenshots (optional)

image

### DIFF
--- a/docs/maps/maps.mdx
+++ b/docs/maps/maps.mdx
@@ -953,7 +953,7 @@ export default RadarMap;
  </TabItem>
 </Tabs>
 
-### Display a Polygon
+### Display a polygon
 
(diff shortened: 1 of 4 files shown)


boonboonsiri (changes requested):
I don't think the globe actually renders. I think we need to update the version of maplibre-gl for our docs site.

Not sure which version it was added but maplibre gl's docs use

```
npm install maplibre-gl@5.3.0
```

PR #607, opened 2025-03-27
support more nesting in nav

kochis:
* Support more nesting in the righthand nav


kochis:
I guess this is not supported in our version of Docusaurus

PR #623, opened 2025-07-18
Historic Traffic DateTime parameter to Route Endpoints

boonboonsiri:

### DIFF
--- a/docs/api.mdx
+++ b/docs/api.mdx
@@ -642,9 +642,9 @@ Autocompletes partial addresses and place names, sorted by relevance and proximi
 - **`lang`** (string, optional): Specifies the language for the results. A string, one of `ar`, `de`, `en`, `es`, `fr`, `ja`, `ko`, `pt`, `ru`, `zh`. Defaults to `en`.
 
 ###### Query best practices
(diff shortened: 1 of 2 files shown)


docs/api.mdx
```diff
@@ -985,6 +985,7 @@ Calculates the travel distance and duration between an origin and a destination.
 - **`units`** (string, optional): The distance units. A string, `metric` or `imperial`. Defaults to `imperial` if not provided. In the response, `distance.value` will be in meters for `metric` and in feet for `imperial`.
 - **`avoid`** (string, optional): The features that calculated routes should avoid. A string, comma-separated, including one or more of `tolls`, `highways`, `ferries` and `borderCrossings`. No features are avoided if not provided.
 - **`geometry`** (string, optional): The format of the `geometry` in the response.  Valid values are `linestring`, `polyline5` and `polyline6`. `linestring` returns a GeoJSON `LineString`, `polyline5` returns a polyline with 5 decimal places of precision (compatible with other mapping providers) and `polyline6` returns a polyline with 6 decimal places of precision. Defaults to none if not provided.
+- ** `dateTime`** (string, optional): The date and time of the trip provided as a string in `YYYY-MM-DDTHH:MM` format which will be used for historical traffic times. If not provided, the current date and time will be used.
```
bschoeneweis, line 988: only nit here would be if we want to use the term `trip` to not conflate with the trip endpoints

